PR #144, opened 2021-09-15
Adding MPU6886 Gyroscope & Accelerometer

jantielens:
## Description

- Implementing the MPU6886 which is used in the M5StackC Plus board
- Reads gyroscope, accelerometer and internal temperature
- Including basic sample & docs

## Motivation and Context

- Was not yet available 😁

## How Has This Been Tested?

Tested on a real M5StickC Plus board, data is similar to the Arduino sample provided by M5Stack.

## Screenshots
image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Unit Tests (work on Unit Tests, has no impact on code or features)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.


devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,176 @@
...
+namespace Iot.Device.Mpu6886
+{
+    /// <summary>
+    /// Mpu6886 accelerometer and gyroscope
+    /// </summary>
+    [Interface("Mpu6886 accelerometer and gyroscope")]
+    public class Mpu6886AccelerometerGyroscope : IDisposable
+    {
+        private I2cDevice _i2c;
+        private const double gyroscopeResolution = 2000.0 / 32768.0;
```
Ellerbach, line 21: For the naming conventions, please use Capital letter for constants. So in this case 'GyroscopeResolution'. Same for the next one.
A good practice as well is to place a little remark on where those constants may come from.

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,176 @@
...
+            I2cDevice i2cDevice)
+        {
+            _i2c = i2cDevice ?? throw new ArgumentNullException(nameof(i2cDevice));
+
+            SpanByte readBuffer = new byte[1];
+
+            _i2c.WriteByte((byte)Mpu6886.Register.WhoAmI);
+            _i2c.Read(readBuffer);
+            if (readBuffer[0] != 0x19)
+                throw new IOException($"This device does not contain the correct signature 0x19 for a MPU6886");
```
Ellerbach, line 36: We are using { } for all if even single line ones. That's valid everywhere for all if, else and if else.
Please adjust all this every where there may be.
The reason behind is to reduce errors.

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,176 @@
...
+
+            _i2c.Write(new SpanByte(new byte[] { (byte)Mpu6886.Register.PowerManagement1, 0x00 }));
+            Thread.Sleep(10);
+
+            _i2c.Write(new SpanByte(new byte[] { (byte)Mpu6886.Register.PowerManagement1, 0x01 << 7 }));
```
Ellerbach, line 41: rather than 0x01 << 7 just directly use the right 0b01000_0000. It's much easier to read and understand or directly 0x80. 
In most cases doing the << doesn't bring any value and do not bring clarity neither for constants.

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,176 @@
...
+            _i2c.Write(new SpanByte(new byte[] { (byte)Mpu6886.Register.AccelerometerConfiguration1, 2 << 3 }));
+            Thread.Sleep(10);
+        }
+
+        private Vector3 GetRawAccelerometer()
+        {
+            SpanByte vec = new byte[6];
+            Read(Mpu6886.Register.AccelerometerMeasurementXHighByte, vec);
+
+            var x = (vec[0] << 8 | vec[1]);
```
Ellerbach, line 100: Optional: BinaryPrimitives are here for this. But leave it like that, it's ok.

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,176 @@
...
+        /// Reads the current gyroscope values from the registers, and compensates them with the gyroscope resolution.
+        /// </summary>
+        /// <returns>Vector of the rotation</returns>
+        public Vector3 GetGyroscope() => GetRawGyroscope() * gyroscopeResolution;
+
+        /// <summary>
+        /// Reads the register of the on-chip temperature sensor which represents the MPU-6886 die temperature.
+        /// </summary>
+        /// <returns>Temperature in degrees Celcius</returns>
+        public double GetTemperature() => GetRawTemperature() / 326.8 + 25.0; // p43 of datasheet describes the room temp. compensation calcuation
```
Ellerbach, line 143: Please use UnitsNet for Temperature. So anyone can then consume it the way they like.

devices/Mpu6886/Mpu6886.sln
```diff
@@ -0,0 +1,56 @@
...
+Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "System.Device.Model", "..\..\src\System.Device.Model\System.Device.Model.shproj", "{23325B14-3651-4879-9697-9846FF123FEB}"
+EndProject
+Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "System.Numerics", "..\..\src\System.Numerics\System.Numerics.shproj", "{F5CCB96C-70B1-4F0A-9F0E-BB912B6A996C}"
+EndProject
+Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "Stopwatch", "..\..\src\Stopwatch\Stopwatch.shproj", "{875D133B-033B-4D11-B1A2-DEF898349AD5}"
+EndProject
+Global
+	GlobalSection(SharedMSBuildProjectFiles) = preSolution
+		..\..\src\System.Device.Model\System.Device.Model.projitems*{23325b14-3651-4879-9697-9846ff123feb}*SharedItemsImports = 13
+		..\..\src\BinaryPrimitives\BinaryPrimitives.projitems*{3f28b003-6318-4e21-a9b6-6c0dbd0bdbfd}*SharedItemsImports = 13
```
Ellerbach, line 22: we do have nugets for all those now. You can use the nuget tather than the shared project.

devices/Mpu6886/README.md
```diff
@@ -0,0 +1,34 @@
...
+// the M5StickC Plus. These pins might be different for other boards.
+
+Configuration.SetPinFunction(22, DeviceFunction.I2C1_CLOCK);
+Configuration.SetPinFunction(21, DeviceFunction.I2C1_DATA);
+
+I2cConnectionSettings settings = new(1, 0x68);
+
+using (Mpu6886AccelerometerGyroscope ag = new(I2cDevice.Create(settings)))
+{
+    Debug.WriteLine($"Internal temperature: {ag.GetTemperature()} C");
```
Ellerbach, line 21: That will need to be adjusted when you'll have the Temperature UnitsNet

devices/Mpu6886/samples/Program.cs
```diff
@@ -0,0 +1,38 @@
...
+            // I2C pins need to be configured, for example for pin 22 & 21 for 
+            // the M5StickC Plus. These pins might be different for other boards.
+            Configuration.SetPinFunction(22, DeviceFunction.I2C1_CLOCK);
+            Configuration.SetPinFunction(21, DeviceFunction.I2C1_DATA);
+
+            I2cConnectionSettings settings = new(1, 0x68);
+
+            using (Mpu6886AccelerometerGyroscope ag = new(I2cDevice.Create(settings)))
+            {
+                Debug.WriteLine($"Internal temperature: {ag.GetTemperature()} C");
```
Ellerbach, line 25: Same here, will need to be adjusted once you'll use UnitsNet

devices/Mpu6886/samples/README.md
```diff
@@ -0,0 +1,51 @@
+﻿# Mpu6886 - accelerometer and gyroscope
```
Ellerbach, line 1: please remove this file, not needed, only the one in the /Mpu6886 is needed

Ellerbach (commented):
Thanks, few adjustment needed and on top:

- Please add the Nerdbanck git versionning nuget. Check out the other projects.
- You'll need to adjust the nfproj
- You need to add a nuspec file as well

You have examples in all the other bindings.

Ellerbach:
Also, please make sure you pull the latest changes in the repository, you can do it from here and then pull them in your branch.
For the temperature, are you sure about the math? 64°C seems very high!

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,176 @@
...
+            _i2c.Write(new SpanByte(new byte[] { (byte)Mpu6886.Register.PowerManagement1, 0x00 }));
+            Thread.Sleep(10);
+
+            _i2c.Write(new SpanByte(new byte[] { (byte)Mpu6886.Register.PowerManagement1, 0x01 << 7 }));
+            Thread.Sleep(10);
+
+            _i2c.Write(new SpanByte(new byte[] { (byte)Mpu6886.Register.PowerManagement1, 0x01 << 0 }));
+            Thread.Sleep(10);
+
+            // ACCEL_CONFIG(0x1c) : +-8G
```
Ellerbach, line 47: You should be able to expose all this in a proper way. The hard coded configuration is very limiting.

devices/Mpu6886/Mpu6886.cs
Ellerbach: Checked, according to the documentation it's -25.0, not +25 ;-)

Ellerbach:
Few more questions:

- I've been looking at the datasheet and it seems that there are compensations registers, you don't use them at all, is there a specific reason why?
- For the long configuration that you are running with magic numbers, it would be much much much better to expose the settings. You can look at the recently merged Axp192 where there is a good mix of properties to change the modes and function to setup properly the modes. As now the way you set it up is just for one purpose and there is no way to change the modes.
- All gypros dps, accelerometer sample rates, all those should be exposed as one size does not fit all
- The low power mode configuration should be accessible as well
- Same for the wake up motion especially on MCU!
- optional: expose the self tests

devices/Mpu6886/Mpu6886.cs
jantielens: Fixed capitalization & added description for constants.

devices/Mpu6886/Mpu6886.cs
jantielens: Unitsnet added. 
Can we double check the offset @Ellerbach ? I know Arduino libraries are not the best to look at, but the [M5 Stack library also does +25](https://github.com/m5stack/M5StickC/blob/3e00ecfa0897a432995a3f80874715cbe6ad60ee/src/utility/MPU6886.cpp#L336). And for me the datasheet is not super clear on this. Happy to follow your lead, if you are still convinced. :)

devices/Mpu6886/Mpu6886.sln
jantielens: Makes sense, so adjusting. 
@Ellerbach FYI [there are other device .sln files in the repo](https://github.com/nanoframework/nanoFramework.IoT.Device/search?q=23325b14-3651-4879-9697-9846ff123feb) that still use shared projects. Might need some cleanup later on.

devices/Mpu6886/Mpu6886.cs
Ellerbach: Run it and if the room temperature is about the same as you read, then offset is in the right direction. 64°C is clearly not the right temperature for a sensor like this. So I would personally give it a try with the -25C and see how it goes.

devices/Mpu6886/Mpu6886.sln
Ellerbach: I know and they have to be adjusted over time. It's just a lot of work to do. While for the new ones, we can directly move to nuget.

devices/Mpu6886/Mpu6886.cs
jantielens: Switching to -25 results in temperature of 13C (even for Belgium that's cold 😁). So I'll leave the calculation as is. The datasheet says it's the chip's "die temperature", and my device feels pretty warm in hand, so I'm willing to believe the die temp is high.

Ellerbach: Very fair enough!

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,371 @@
...
+
+            return offset;
+        }
+
+        /// <summary>
+        /// Write the gyroscope offset in the GyroscopeOffsetAdjustment registers of the MPU6886.
+        /// This function can be usefull when a custom callibration calculation is used, instead of the Calibrate function of this class.
+        /// </summary>
+        /// <param name="offset">The vector containing the offsets for the 3 axes.</param>
+        public void SetGyroscopeOffset(Vector3 offset)
```
Ellerbach, line 211: As you have a Get and a Set for this one and as it doesn't change over time, you can easily transform it as a property. Just rename those 2 functions as private and Create a GyroscopeOffet property calling those 2 private functions. This will be even nicer!

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,371 @@
...
+        public void Reset()
+        {
+            _i2c.Write(new SpanByte(new byte[] { (byte)Mpu6886.Register.PowerManagement1, 0b1000_0000 }));
+            Thread.Sleep(10);
+        }
+
+        /// <summary>
+        /// Set the chip to sleep mode. (Datasheet, page 47)
+        /// </summary>
+        public void Sleep()
```
Ellerbach, line 262: You may want to add the wakeup function as well ;-)

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,371 @@
...
+        {
+            _i2c.Write(new SpanByte(new byte[] { (byte)Mpu6886.Register.PowerManagement1, 0b0100_0000 }));
+            Thread.Sleep(10);
+        }
+
+        /// <summary>
+        /// Sets the accelerometer full scale. (Datasheet page 37)
+        /// </summary>
+        /// <param name="scale">The scale to set.</param>
+        public void SetAccelerometerScale(AccelerometerScale scale)
```
Ellerbach, line 272: Same as for the Gyroscope, this is a great candidate for a property

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,371 @@
...
+        {
+            _i2c.Write(new SpanByte(new byte[] { (byte)Mpu6886.Register.AccelerometerConfiguration1, (byte)scale }));
+            Thread.Sleep(1);
+        }
+
+        /// <summary>
+        /// Sets the gyroscope full scale. (Datasheet page 37)
+        /// </summary>
+        /// <param name="scale">The scale to set.</param>
+        public void SetGyroscopeScale(GyroscopeScale scale)
```
Ellerbach, line 282: And this one too, can be done as a property.

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,371 @@
...
+        {
+            get
+            {
+                SpanByte readBuffer = new byte[1];
+                _i2c.WriteByte((byte)Mpu6886.Register.PowerManagement2);
+                _i2c.Read(readBuffer);
+
+                // bit 1 in the register means disabled, so using bitwise not to flip bits.
+                return (EnabledAxis)~readBuffer[0];
+            }
```
Ellerbach, line 301: missing an extra enter

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,371 @@
...
+                _i2c.Write(new SpanByte(new byte[] { (byte)Mpu6886.Register.PowerManagement2, (byte)~value }));
+                Thread.Sleep(1);
+            }
+        }
+
+        /// <summary>
+        /// Sets the averaging filter settings for low power accelerometer mode. (Datasheet page 37)
+        /// </summary>
+        /// <param name="mode">The mode to set.</param>
+        public void SetAcceleratorLowPowerMode(AccelerometerLowPowerMode mode)
```
Ellerbach, line 314: As you can as well read the AccelerometerLowPowerMode, I would transform this one as well as a property

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,371 @@
...
+        /// </summary>
+        public byte SampleRateDivider
+        {
+            get
+            {
+                SpanByte readbuffer = new byte[1];
+                _i2c.WriteByte((byte)Mpu6886.Register.SampleRateDevider);
+                _i2c.Read(readbuffer);
+                return readbuffer[0];
+            }
```
Ellerbach, line 343: missing an extra enter

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,371 @@
...
+        /// </summary>
+        public InterruptEnable AccelerometerInterruptEnabled
+        {
+            get
+            {
+                SpanByte readbuffer = new byte[1];
+                _i2c.WriteByte((byte)Mpu6886.Register.InteruptEnable);
+                _i2c.Read(readbuffer);
+                return (InterruptEnable)readbuffer[0];
+            }
```
Ellerbach, line 362: missing an extra enter

devices/Mpu6886/Mpu6886.nfproj
```diff
@@ -0,0 +1,79 @@
...
+  <ItemGroup>
+    <Compile Include="Properties\AssemblyInfo.cs" />
+    <None Include="*.md" />
+  </ItemGroup>
+  <ItemGroup>
+    <Content Include="category.txt" />
+    <Content Include="version.json" />
+  </ItemGroup>
+  <Import Project="$(NanoFrameworkProjectSystemPath)NFProjectSystem.CSharp.targets" Condition="Exists('$(NanoFrameworkProjectSystemPath)NFProjectSystem.CSharp.targets')" />
+  <Import Project="..\..\src\BinaryPrimitives\BinaryPrimitives.projitems" Label="Shared" />
```
Ellerbach, line 64: you still have shared project here. Please use the nugets instead.

Ellerbach: I guess this file may not have been push properly as you did the nuspec properly :-)

devices/Mpu6886/Mpu6886.sln
```diff
@@ -0,0 +1,36 @@
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.30413.136
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{11A8DD76-328B-46DF-9F39-F559912D0360}") = "Mpu6886", "Mpu6886.nfproj", "{E79903D4-4F7C-4FB5-8078-086CD3B3252B}"
+EndProject
+Project("{11A8DD76-328B-46DF-9F39-F559912D0360}") = "Mpu6886.Samples", "samples\Mpu6886.Samples.nfproj", "{7EC388BC-55BB-4506-B93B-AD931F3F6D80}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Shared Projects", "Shared Projects", "{E785CAF1-60D9-4191-9AFB-C49027D3233A}"
```
Ellerbach, line 9: you may remove this one as empty now

devices/Mpu6886/README.md
```diff
@@ -0,0 +1,106 @@
+﻿# Mpu6886 - accelerometer and gyroscope
+
```
Ellerbach, line 2: maybe add few words on what it is

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,371 @@
...
+
+namespace Iot.Device.Mpu6886
+{
+    /// <summary>
+    /// Mpu6886 accelerometer and gyroscope
+    /// </summary>
+    [Interface("Mpu6886 accelerometer and gyroscope")]
+    public class Mpu6886AccelerometerGyroscope : IDisposable
+    {
+        private I2cDevice _i2c;
```
Ellerbach, line 21: Please add a `public const int DefaultI2cAddress = 0x68` with a proper description. This makes life easier for the developers.
Note, public constants goe first and require a /// comment.

devices/Mpu6886/README.md
```diff
@@ -0,0 +1,106 @@
...
+⚠ **When the devices reboots, the offset registers are cleared.** So if you don't persist the calibration data yourself
+you will have to run the calibration method every time the device boots.
+
+## Sleep mode
+
+The sensor can be put in sleep mode by calling the `Sleep` function.
+
+## Setting scales
+
+Both for the gyroscope and accelerometer you can set the desired scales using the `SetAccelerometerScale` and `SetGyroscopeScale`
```
Ellerbach, line 89: See my comment in the code, those are great candidates for properties, don't forget to adjust the doc as well.

devices/Mpu6886/samples/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,10 @@
+using System.Reflection;
```
Ellerbach, line 1: missing header

Ellerbach (commented):
Thanks a lot! Almost there!

jantielens:
@Ellerbach thanks for all the feedback. ♥ Addressed all of it, so feel free to go over the PR again.

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,419 @@
...
+        /// <summary>
+        /// Gets and sets the accelerometer full scale. (Datasheet page 37)
+        /// </summary>
+        public AccelerometerScale AccelerometerScale
+        {
+            get
+            {
+                SpanByte buffer = new byte[1];
+                Read(Mpu6886.Register.AccelerometerConfiguration1, buffer);
+                return (AccelerometerScale)buffer[0];
```
Ellerbach, line 290: Are you sure it's the only element in this register? In all cases, better to adjust like:
```
return (AccelerometerScale)(buffer[0] & 0b0001_1000);
```

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,419 @@
...
+            }
+
+            set
+            {
+                _i2c.Write(new SpanByte(new byte[] { (byte)Mpu6886.Register.AccelerometerConfiguration1, (byte)value }));
```
Ellerbach, line 295: if this is not the only element to setup in this register, you should first read the register, clear the bits related to what you want to setup and then apply the changes.
Similar as you did for AccelerometerLowPowerMode

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,419 @@
...
+                Thread.Sleep(1);
+            }
+        }
+
+        /// <summary>
+        /// Gets and sets the gyroscope full scale. (Datasheet page 37)
+        /// </summary>
+        public GyroscopeScale GyroscopeScale
```
Ellerbach, line 303: Similar question as the previous one

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,419 @@
...
+                _i2c.Write(new SpanByte(new byte[] { (byte)Mpu6886.Register.GyroscopeConfiguration, (byte)value }));
+                Thread.Sleep(1);
+            }
+        }
+
+
+        /// <summary>
+        /// Sets the enabled axes of the gyroscope and accelerometer. (Datasheet page 47)
+        /// </summary>
+        public EnabledAxis EnabledAxes
```
Ellerbach, line 323: Similar question as the previous one

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,419 @@
...
+            {
+                _i2c.Write(new SpanByte(new byte[] { (byte)Mpu6886.Register.SampleRateDevider, value }));
+                Thread.Sleep(1);
+            }
+        }
+
+        /// <summary>
+        /// The axes on which the interrupt should be enabled.
+        /// </summary>
+        public InterruptEnable AccelerometerInterruptEnabled
```
Ellerbach, line 401: similar question here as well

devices/Mpu6886/README.md
```diff
@@ -0,0 +1,111 @@
...
+## Setting enabled axes
+
+By default all gyroscope and accelerometer axes are enabled. If desired you can specify explicitly which axes should be enabled.
+In that case, all other axes will be disabled. When an axis is disabled, data can still be read for that axis but the values will not
+change anymore.
+
+```csharp
+// Enable only the X axis of the gyroscope and accelerometer
+ag.EnabledAxes = EnabledAxis.GyroscopeX | EnabledAxis.AccelerometerX;
+```
```
Ellerbach, line 111: Would be great if you can add at the end of the readme a list of what's not implemented from the sensor.

Ellerbach (commented):
Looks very good, few open questions and then we'll be good to go!

devices/Mpu6886/Mpu6886.cs
jantielens: It's only the 3 bits to start the self test which are in the register (together with the scale). So potentially when the self test would be implemented, this could become an issue. Easy enough to avoid, so adjusting now.

devices/Mpu6886/Mpu6886.cs
jantielens: See previous comment, no problem now but avoid in the future I'll adjust.

devices/Mpu6886/Mpu6886.cs
jantielens: Fixed to avoid potential future issues.

devices/Mpu6886/Mpu6886.cs
jantielens: 2 reserved (unused?) bits in the register, but added a mask just to on the safe side.

jantielens:
@Ellerbach will this be the one? 😂 (thanks again!)

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,466 @@
...
+    /// </summary>
+    [Interface("Mpu6886 accelerometer and gyroscope")]
+    public class Mpu6886AccelerometerGyroscope : IDisposable
+    {
+        /// <summary>
+        /// The default I2C address for the MPU6886 sensor. (Datasheet page 49)
+        /// Mind that the address can be configured as well for 0x69 depending upon the value driven on AD0 pin.
+        /// </summary>
+        public const int DefaultI2cAddress = 0x68;
+
```
Ellerbach, line 26:
```suggestion

        /// <summary>
        /// The secondary I2C address for the MPU6886 sensor. (Datasheet page 49)
        /// </summary>
        public const int SecondaryI2cAddress = 0x69 ;

```

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,466 @@
...
+            {
+                // First read the current register values
+                SpanByte currentRegisterValues = new byte[1];
+                _i2c.WriteByte((byte)Mpu6886.Register.AccelerometerConfiguration1);
+                _i2c.Read(currentRegisterValues);
+
+                byte mask = 0b1110_0111; // we leave all bits except bit 3 and 4 untouched with this mask
+                byte cleaned = (byte)(currentRegisterValues[0] & mask);
+
+                byte newvalue = (byte)(cleaned | (byte)value); // apply the new scale
```
Ellerbach, line 303: To avoid having to allocate 2 bytes which are not need, I would use a condensed version like this:

```suggestion
                byte newvalue = (byte)((currentRegisterValues[0] & 0b1110_0111) | (byte)value); // apply the new scale
```

Please apply this else where. It's important to not allocate anything when not needed.

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,466 @@
...
+            }
+
+            set
+            {
+                // First read the current register values
+                SpanByte currentRegisterValues = new byte[1];
+                _i2c.WriteByte((byte)Mpu6886.Register.GyroscopeConfiguration);
+                _i2c.Read(currentRegisterValues);
+
+                byte mask = 0b1110_0111; // we leave all bits except bit 3 and 4 untouched with this mask
```
Ellerbach, line 331: See previous comment about allocation

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,466 @@
...
+            }
+
+            set
+            {
+                // First read the current register values
+                SpanByte currentRegisterValues = new byte[1];
+                _i2c.WriteByte((byte)Mpu6886.Register.PowerManagement2);
+                _i2c.Read(currentRegisterValues);
+
+                byte mask = 0b1100_0000; // we leave all bits except bit 7 and 6 untouched with this mask
```
Ellerbach, line 366: Same, see previous comment about allocation

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,466 @@
...
+        /// Gets and sets the averaging filter settings for low power accelerometer mode. (Datasheet page 37)
+        /// </summary>
+        public AccelerometerLowPowerMode AccelerometerLowPowerMode
+        {
+            get {
+                SpanByte currentRegisterValues = new byte[1];
+                _i2c.WriteByte((byte)Mpu6886.Register.AccelerometerConfiguration2);
+                _i2c.Read(currentRegisterValues);
+
+                byte mask = 0b0011_0000; // we leave all bits except bit 4 and 5 untouched with this mask
```
Ellerbach, line 389: see previous comment about allocation

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,466 @@
...
+            }
+
+            set
+            {
+                // First read the current register values
+                SpanByte currentRegisterValues = new byte[1];
+                _i2c.WriteByte((byte)Mpu6886.Register.AccelerometerConfiguration2);
+                _i2c.Read(currentRegisterValues);
+
+                byte mask = 0b1100_1111; // we leave all bits except bit 4 and 5 untouched with this mask
```
Ellerbach, line 402: +1

devices/Mpu6886/Mpu6886.cs
```diff
@@ -0,0 +1,466 @@
...
+            }
+            
+            set
+            {
+                // First read the current register values
+                SpanByte currentRegisterValues = new byte[1];
+                _i2c.WriteByte((byte)Mpu6886.Register.InteruptEnable);
+                _i2c.Read(currentRegisterValues);
+
+                byte mask = 0b0011_1111; // we leave all bits except bit 4 and 5 untouched with this mask
```
Ellerbach, line 454: And this one too :-)

devices/Mpu6886/README.md
```diff
@@ -0,0 +1,128 @@
...
+- register `0x1a`: FIFO mode, DLPF (digital low pass filter)
+- register `0x1d` bit 0 to 3: DLPF
+- register `0x1e` bit 4 to 6: averaging filter for low power mode
+- registers `0x20` to `0x22`: threshold for wake-on motion
+- register `0x36`: fsync interrupt status
+- register `0x37`: INT/DRDY pin config
+- register `0x39` and `0x3a`: FIFO watermark interrupt
+- register `0x60` and `0x61`: FIFO watermark interrupt
+- register `0x68`: signal path reset
+- register `0x69`: acc. intelligence control
```
Ellerbach, line 128: missing an extra enter at the end (to be Markdown perfect ;-))

Ellerbach (commented):
One more adjustment and it will be perfect!

devices/Mpu6886/README.md
jantielens: Already there, so leaving this one :)

devices/Mpu6886/Mpu6886.cs
Ellerbach: Don't you want to add a secondary I2C address as suggested?

Ellerbach (approved):
Looks good to me! @josesimoes you may want to have a look to check if anything is missing or needs adjustmeent.

devices/Mpu6886/Mpu6886.nfproj
```diff
@@ -0,0 +1,91 @@
...
+    <RootNamespace>Iot.Device.Mpu6886</RootNamespace>
+    <AssemblyName>Iot.Device.Mpu6886</AssemblyName>
+    <TargetFrameworkVersion>v1.0</TargetFrameworkVersion>
+    <DocumentationFile>bin\$(Configuration)\Iot.Device.Mpu6886.xml</DocumentationFile>
+    <LangVersion>9.0</LangVersion>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+  </PropertyGroup>
+  <Import Project="$(NanoFrameworkProjectSystemPath)NFProjectSystem.props" Condition="Exists('$(NanoFrameworkProjectSystemPath)NFProjectSystem.props')" />
+  <ItemGroup>
+    <Reference Include="mscorlib">
```
josesimoes, line 24: This should be like this, otherwise the updates will bite us.

```suggestion
    <Reference Include="mscorlib, Version=1.10.5.4, Culture=neutral, PublicKeyToken=c07d481e9758c731">
```

josesimoes (approved):
All good, please just update that mscrolib ref and we're good to merge! 👏🏻

josesimoes (approved):
👌🏻

Ellerbach:
@jantielens huge thanks fort this super nice binding! Thanks for the efforts and going thru all feedbacks!

PR #172, opened 2021-10-11
Updated the OLED SSD1306 Demos and Drivers with Basic Functionality t…

gligorov:
…o work with latest nF

Updated the code to make minimal changes to make it work with OLED. Tested with 128x64 and 128x32 and it works fine.

Updated Font logic and created basic Font logic and drawing on screen.

There is more work to do, but this is initial update to make the code work, and see what direction will need to be taken for the other controllers/devices if they will be supported.

## Motivation and Context

- Fixes/Closes/Resolves nanoFramework/Home#NNNN

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Unit Tests (work on Unit Tests, has no impact on code or features)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.


MichaelWheeley:
I tried your code but had no success with my 128x32 device. Debugging in detail I noticed that in the Ssd13xx.cs display method around line 251 you have this code,
_pageCmd[1] = (byte)(PageAddress.Page0 + i); // page number
Now in the enum Page0 has value zero, he page address should be in range 0xb0 -> 0xb7 so line 251 appears to need a 0xb0 offset
i.e. _pageCmd[1] = (byte)(0xb0 + PageAddress.Page0 + i); // page number
and only after making modification works nicely
Capture

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -101,5 +142,172 @@ protected SpanByte SliceGenericBuffer(int start, int length)
...
         }
+
+
+        /// <summary>
+        /// Draws a pixel on the screen.
+        /// </summary>
+        /// <param name="x">The x coordinate on the screen.</param>
+        /// <param name="y">The y coordinate on the screen.</param>
+        /// <param name="useWhite">Indicates if color to be used, default yes unless is inverted screen.</param>
+        public void DrawPixel(int x, int y, bool useWhite = true)
```
Ellerbach, line 153: I feel like `useWhite` should be renamed to `inverted` because at the end of the day, it's about that.

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -101,5 +142,172 @@ protected SpanByte SliceGenericBuffer(int start, int length)
...
+        {
+            if ((x >= Width) || (y >= Height))
+                return;
```
Ellerbach, line 156: We're using brackets { } around all instructions for if and kind of loops. So please add. This applies in other places. I'll try to point them out and will use the word brackets.

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -101,5 +142,172 @@ protected SpanByte SliceGenericBuffer(int start, int length)
...
+
+            // x specifies the column
+            int idx = x + (y / 8) * Width;
+
+            if (useWhite)
+                _genericBuffer[idx] |= (byte)(1 << (y & 7));
```
Ellerbach, line 162: brackets

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -101,5 +142,172 @@ protected SpanByte SliceGenericBuffer(int start, int length)
...
+            else
+                _genericBuffer[idx] &= (byte)~(1 << (y & 7));
+
+            //INVERSE COLOR: _genericBuffer[idx] ^= (byte)(1 << (y & 7));
```
Ellerbach, line 166: please remove any dead code that is not use. If you really want to keep dead code, please add a comment why. This helps getting everything in a clean state

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -101,5 +142,172 @@ protected SpanByte SliceGenericBuffer(int start, int length)
...
+        }
+
```
Ellerbach, line 168: extra carrer return

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -101,5 +142,172 @@ protected SpanByte SliceGenericBuffer(int start, int length)
...
+        /// </summary>
+        /// <param name="x"></param>
+        /// <param name="y"></param>
+        /// <param name="width"></param>
+        /// <param name="height"></param>
+        /// <param name="bitmap"></param>
+        public void DrawBitmap(int x, int y, int width, int height, byte[] bitmap)
+        {
+            if ((width * height) != bitmap.Length)
+                throw new ArgumentException("Width and height do not match the bitmap size.");
```
Ellerbach, line 181: brackets

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -101,5 +142,172 @@ protected SpanByte SliceGenericBuffer(int start, int length)
...
+                    {
+                        DrawPixel(x + (8 * xA) + pixel, y + yO, (b & mask) > 0);
+                        mask <<= 1;
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        ///
```
Ellerbach, line 199: Missing comments

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -101,5 +142,172 @@ protected SpanByte SliceGenericBuffer(int start, int length)
...
+        {
+            if ((width * height) != bitmap.Length)
+                throw new ArgumentException("Width and height do not match the bitmap size.");
+
+            for (var yO = 0; yO < height; yO++)
+            {
+                for (var xA = 0; xA < width; xA++)
+                {
+                    var b = bitmap[(yO * width) + xA];
+                    byte mask = 0x01;
```
Ellerbach, line 188: you may define the byte definition outside of the loop and just use mask = 0x01; here

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -101,5 +142,172 @@ protected SpanByte SliceGenericBuffer(int start, int length)
...
+
+        /// <summary>
+        /// Displays the information on the screen.
+        /// </summary>
+        public void Display()
+        {
+            for (byte i = 0; i < Pages; i++)
+            {
+                _pageCmd[1] = (byte)(PageAddress.Page0 + i); // page number
+                _i2cDevice.Write(_pageCmd);
```
Ellerbach, line 251: See the comment in the PR, seems like this part may need to be adjusted for some screens

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -101,5 +142,172 @@ protected SpanByte SliceGenericBuffer(int start, int length)
...
+        ///     Sequence of bytes that should be sent to a 128x64 OLED display to setup the device.
+        ///		First byte is the command byte 0x00.
+        /// </summary>
+        private readonly byte[] _oled128x64Init =
+        {
+            0x00, 0xae, 0xd5, 0x80, 0xa8, 0x3f, 0xd3, 0x00, 0x40 | 0x0, 0x8d, 0x14, 0x20, 0x00, 0xa0 | 0x1, 0xc8,
+            0xda, 0x12, 0x81, 0xcf, 0xd9, 0xf1, 0xdb, 0x40, 0xa4, 0xa6, 0xaf
+        };
+
+        /// <summary>
```
Ellerbach, line 281: please adjust the comments by removing the extra spaces
And also a space alignement problem. In Visual Studio, you can use ctrl + k + d
This automatically correct those kind of miss alignements.

devices/Ssd13xx/samples/Program.cs
```diff
@@ -1,287 +1,23 @@
...
 
-                return address.Address.ToString();
-            }
-        }
-    }
 
-    return null;
-}
+//Tested with 128x64 and 128x32 OLEDs
+using Ssd1306 device = new Ssd1306(I2cDevice.Create(new I2cConnectionSettings(2, 0x3C)), Ssd13xx.DisplayResolution.OLED128x64);
```
Ellerbach, line 17: For the I2C address, a good pattern is to add in the Ssd1306 class something like: public const int DefaultI2cAddress = 0x3C;

Ellerbach: Also as this is using I2C, please look at any other sample using I2C, there is a special commented part to add at the beginning of the code to help people understand that they need to setup properly the pins if needed. See for example: https://github.com/nanoframework/nanoFramework.IoT.Device/blob/04ac94cefb39f70312ada24b12ae7cb289cbe219/devices/Bmxx80/samples/Bmp280.sample.cs#L26

devices/Ssd13xx/Ssd1306.cs
```diff
@@ -19,11 +19,22 @@ public class Ssd1306 : Ssd13xx
...
+        /// <summary>
+        /// Initializes new instance of Ssd1306 device that will communicate using I2C bus.
+        /// A single-chip CMOS OLED/PLED driver with controller for organic/polymer
+        /// light emitting diode dot-matrix graphic display system.
+        /// </summary>
+        /// <param name="i2cDevice">The I2C device used for communication.</param>
+        /// <param name="res">Display resolution</param>
+        public Ssd1306(I2cDevice i2cDevice, DisplayResolution res) : base(i2cDevice, res)
+        {
+
```
Ellerbach, line 35: extra enter

Ellerbach (commented):
Thanks a lot, this is huge step forward to have those small screens useful with our lovely .NET nanoFramework devices :-) I have few but don't have them right now with me. I'll test this asap as well!
I left some comments, they are mainly about cosmetics, let me know if they are not clear. 
Also, the font file is missing.

devices/Ssd13xx/Ssd13xx.cs
MichaelWheeley: ty, that works

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -101,5 +142,172 @@ protected SpanByte SliceGenericBuffer(int start, int length)
...
+            }
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="x"></param>
+        /// <param name="y"></param>
+        /// <param name="str"></param>
+        public void DrawString(int x, int y, string str)
```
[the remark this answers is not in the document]
MichaelWheeley: There is a fundamental difference between DrawString used here and Write seen in other similar drivers, the x and y coordinates are scaled in pixels. The difficulty with this approach is that the user must know the font size and calculate the position.
Might I propose that you keep DrawString exactly as-is by correct the comments to say x,y are pixel coordinates,
and create an additional function Write with the same arguments but which scales the x,y into text position coordinates ?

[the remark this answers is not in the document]
MichaelWheeley: something like this,
        /// <summary>
        /// Writes a text message on the screen with font in use.
        /// </summary>
        /// <param name="x">The x pixel-coordinate on the screen.</param>
        /// <param name="y">The y pixel-coordinate on the screen.</param>
        /// <param name="str">Text string to display.</param>
        /// <seealso cref="Write"/>
        public void DrawString(int x, int y, string str)
        {
            byte[] bitMap = GetTextBytes(str);
            DrawBitmap(x, y, bitMap.Length / Font.Height, Font.Height, bitMap);
        }

        /// <summary>
        /// Writes a text message on the screen with font in use.
        /// </summary>
        /// <param name="x">The x text-coordinate on the screen.</param>
        /// <param name="y">The y text-coordinate on the screen.</param>
        /// <param name="str">Text string to display.</param>
        /// <seealso cref="DrawString"/>
        public void Write(int x, int y, string str)
        {
            byte[] bitMap = GetTextBytes(str);
            DrawBitmap((x * this.Font.Width), (y * this.Font.Height), (bitMap.Length / Font.Height), Font.Height, bitMap);
        }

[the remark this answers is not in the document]
gligorov: Added the updates, will see how the PR is merged. At this time I do not have a good method for large text display.  I have some working code that uses several methods and uses DrawRectangle vs DrawPixel to make larger size fonts.  It may be too slow for some boards.  So any ideas suggestions are welcome in that area.

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -101,5 +149,288 @@ protected SpanByte SliceGenericBuffer(int start, int length)
 
             return _genericBuffer;
         }
+
```
Ellerbach, line 152: one extra enter here as well

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -101,5 +149,288 @@ protected SpanByte SliceGenericBuffer(int start, int length)
...
+            if (inverted)
+            {
+                _genericBuffer[idx] |= (byte)(1 << (y & 7));
+            }
+            else
+            {
+                _genericBuffer[idx] &= (byte)~(1 << (y & 7));
+            }
+        }
+
```
Ellerbach, line 179: please remove the extra enter here as well

devices/Ssd13xx/Ssd13xx.cs
Ellerbach: @gligorov you still have this one as well, declare the byte mask; before the first for so it won't get created every time

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -101,5 +149,288 @@ protected SpanByte SliceGenericBuffer(int start, int length)
...
+        /// </summary>
+        /// <param name="x">The x text-coordinate on the screen.</param>
+        /// <param name="y">The y text-coordinate on the screen.</param>
+        /// <param name="str">Text string to display.</param>
+        /// <seealso cref="DrawString"/>
+        public void Write(int x, int y, string str)
+        {
+            this.DrawString(x * this.Font.Width, y * this.Font.Height, str);
+        }
+
```
Ellerbach, line 266: you have an extra enter here as well ;-)

devices/Ssd13xx/Ssd13xx.cs
Ellerbach: @gligorov this is really perfect!!! Thanks a lot!

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -101,5 +149,288 @@ protected SpanByte SliceGenericBuffer(int start, int length)
...
+            /// <summary>
+            /// Option for 128x32 OLED
+            /// </summary>
+            OLED128x32,
+            // <summary>
+            /// Option for 96x16 OLED
+            /// </summary>
+            OLED96x16
+        }
+
```
Ellerbach, line 423: and one more extra enter here. Only 1 needed!

MichaelWheeley:
Capture

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -101,5 +144,323 @@ protected SpanByte SliceGenericBuffer(int start, int length)
...
+        /// Draws a rectangle that is solid/filled.
+        /// </summary>
+        /// <param name="x0">x coordinate starting of the top left.</param>
+        /// <param name="y0">y coordinate starting of the top left.</param>
+        /// <param name="width">Width of rectabgle in pixels.</param>
+        /// <param name="height">Height of rectangle in pixels</param>
+        /// <param name="inverted">Turn the pixel on (true) or off (false).</param>
+        public void DrawFilledRectangle(int x0, int y0, int width, int height, bool inverted = true)
+        {
+            // width--;
```
Ellerbach, line 214: I know I4m a bit picky, but what is the usage of this dead code? Please add a comment why you need to keep it or just remove it. Thanks!

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -101,5 +144,323 @@ protected SpanByte SliceGenericBuffer(int start, int length)
...
+            0xdb,       // set VCOMH deselect level
+            0x40,       // set display start line
+            0xa4,       // set display ON
+            0xa6,       // set normal display
+            0xaf        // turn display on 0xaf
+        };
+
+        /// <summary>
+		/// Sequence of bytes that should be sent to a 128x32 OLED display to setup the device.
+		/// First byte is the command byte 0x00.
```
Ellerbach, line 386: the alignement is still not good here.

PR #417, opened 2022-05-30
Add support for the Sensirion SPS30

martijns:
## Description
* Adds support for the SPS30 Particulate Matter sensor by Sensirion.
* The sensor supports I2C and UART, but this implementation is currently only for the UART interface.
* The `SHDLC` protocol included in this project is Sensirion-specific, but not SPS30-specific. In the future, it could be re-used for other sensors.
* Includes a sample project.

## Motivation and Context
SPS30 support was non-existing.

## How Has This Been Tested?
The driver has been tested on a Pycom WiPy with ESP32 chip with a Sensirion SPS30 attached. The sample project runs successfully on this hardware.

## Types of changes
- [ ] Improvement (non-breaking change that improves a feature, code or algorithm)
- [ ] Bug fix (non-breaking change which fixes an issue with code or algorithm)
- [x] New feature (non-breaking change which adds functionality to code)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Config and build (change in the configuration and build system, has no impact on code or features)
- [ ] Dependencies (update dependencies and changes associated, has no impact on code or features)
- [ ] Unit Tests (add new Unit Test(s) or improved existing one(s), has no impact on code or features)
- [ ] Documentation (changes or updates in the documentation, has no impact on code or features)

## Checklist:
- [x] My code follows the code style of this project (only if there are changes in source code).
- [ ] My changes require an update to the documentation (there are changes that require the docs website to be updated).
- [ ] I have updated the documentation accordingly (the changes require an update on the docs in this repo).
- [x] I have read the [CONTRIBUTING](https://github.com/nanoframework/.github/blob/master/CONTRIBUTING.md) document.
- [x] I have tested everything locally and all new and existing tests passed (only if there are changes in source code).
- [ ] I have added new tests to cover my changes.

W.r.t. checkbox 3: the code is documented, but no other files in the repository have been touched.

PR #433, opened 2022-06-06
Add support for Single in BinaryPrimitives

martijns:
## Description
* Added "Single" (float) support in `System.Buffers.Binary.BinaryPrimitives`
* Added unit tests to make sure it works correctly

## Motivation and Context
Single/float support was missing. Single is used for an upcoming implementation for the Sensirion SPS30, see #417 

## How Has This Been Tested?
* Unittests

## Types of changes
- [x] Unit Tests (add new Unit Test(s) or improved existing one(s), has no impact on code or features)

## Checklist:
- [x] I have added new tests to cover my changes.


devices/System.Buffers.Binary.BinaryPrimitives/BinaryPrimitives.cs
```diff
@@ -206,6 +206,52 @@ public static ulong ReadUInt64LittleEndian(SpanByte source)
...
+        public static float ReadSingleBigEndian(SpanByte source)
+        {
+            if (source.Length < 4)
+            {
+                throw new ArgumentOutOfRangeException();
+            }
+
+            uint ieee754_bits = ReadUInt32BigEndian(source);
+            float flt;
+            unsafe
```
Ellerbach, line 224: nice usage of unsafe.
@josesimoes just to double check, bindings using the nuget won't have to compile unsafe to use this? If so, then I would prefer a way that is safe if possible. Because that may be quite a constraint if we had to change always for unsafe.

Ellerbach (approved):
Looks good to me. Thanks for adding it! And bonus point for the proper tests! Just a question for @josesimoes in the thread regarding unsafe for any code that will use this nuget.

devices/System.Buffers.Binary.BinaryPrimitives/BinaryPrimitives.cs
josesimoes: Not 100% sure on this, but I think this is "contained" to the assembly being compiled. The one referencing it are unaware of this.

Ellerbach: That's why I asked :-D Because I was not 100% sure neither ;-)

martijns: You already have a bunch of projects in `nanoFramework.IoT.Device` that require unsafe code. One example is `nanoFramework.System.Numerics`. When you add that as reference to your project, your own project does not require to allow unsafe either. A bunch of other device packages also use unsafe. So this shouldn't be an issue.

PR #475, opened 2022-06-28
Add EXTENEnable property to AXP192

josesimoes:
## Description
- Add property to control state of EXTEN switch control.

## Motivation and Context

- Need to control this pin with M5Stack TOUGH.

## Types of changes

## Checklist:

- [x] My changes require an update to the documentation (there are changes that require the docs website to be updated).

### DIFF
--- a/devices/Axp192/Axp192.cs
+++ b/devices/Axp192/Axp192.cs
@@ -14,6 +14,7 @@ namespace Iot.Device.Axp192
         private const byte _dcD3SetBit = (1 << 1);
         private const byte _Ldo2SetBit = (1 << 2);
         private const byte _Ldo3SetBit = (1 << 3);
+        private const byte _ExtEnSetBit = (1 << 6);
 
         public const int I2cDefaultAddress = 0x34;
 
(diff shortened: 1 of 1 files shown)


alberk8:
The ExtEn is on 0x10 Address.

image

josesimoes:
@alberk8 thanks for the comment, but from the datasheet I have it looks like EXTEN should be accessed through register 0x12, not 0x10 (see note on table footer).
Also in M5Stack official library they are accessing register 0x12 too, not 0x10, so I'll go with 0x12 until this is proven wrong (as I don't have access to hardware right now to test it).
 
image

alberk8:
> @alberk8 thanks for the comment, but from the datasheet I have it looks like EXTEN should be accessed through register 0x12, not 0x10 (see note on table footer). Also in M5Stack official library they are accessing register 0x12 too, not 0x10, so I'll go with 0x12 until this is proven wrong (as I don't have access to hardware right now to test it).
> 
> image

But When i looked at the M5Stack [code](https://github.com/m5stack/M5Core2/blob/0b905c583003eaccdf81c2f82b235f186903ca69/src/AXP192.cpp#L557) it does point to 0x10.   In fact I have tested it on nF and it seems to work.

josesimoes:
What a mess! 🙄 

The TOUGH code [here](https://github.com/m5stack/M5Tough/blob/215fb2cee54ae351caeb658e4dd51811de969037/src/AXP192.cpp#L573) it's using 0x12...

And the AXP192 datasheet in TOUGH page [here](https://docs.m5stack.com/en/core/tough) (Chinese version only) is rev1.2.
From the comment on the table above (which seems to be added in rev1.2) I believe we can access that control signal from 0x12. 

And the code M5Stack support pointed to JorgeGOgreenYng it's also using 0x12. 

I'll go with 0x12 which is already here and let's wait for a confirmation.

devices/Axp192/Axp192.cs
```diff
@@ -502,6 +503,33 @@ public void EnableDCDC1(bool state)
             I2cWrite(Register.SwitchControleDcDC1_3LDO2_3, buf);
         }
 
+        /// <summary>
+        /// Gets or Sets the state of EXTEN switch control.
+        /// </summary>
+        /// <value>
+        /// <see langword="true"/> if EXTEN switch is enabled, <see langword="false"/> otherwise.
+        /// </value>
+        public bool EXTENEnable
```
Ellerbach, line 512: Not the preper naming :-( Should have been `ExternEnable` or even better `EnableExtern`

Ellerbach:
That's fine for this time but let's try to follow our proper naming pattern also with the acronyms

PR #547, opened 2022-08-18
Ds1302 real-time clock library

rkalwak:
## Description
Ds1302 real-time clock library with StyleCop application on all RTC implementations.

## Motivation and Context

- Extends existing library with Ds1302 implementation

## How Has This Been Tested?

Connected to to ESP32 DevKit1

## Screenshots
image

## Types of changes

## Checklist:


rkalwak:
I have a feeling that would be better and assembly would be tiny if we had separate lib for each implementation which references the common part in places it really needs it. This way if one of four implementation needs a lib it is only attached there and in sample project. I'm open to make it this way.

@Ellerbach

Ellerbach:
@rkalwak, yes, in theory we should break down each RTC into smaller packages. So if you want to make a separate one for this one, feel free to do!
There are dependencies with the base RTC, but give it a try! Happy to start on this. For the existing ones, we can keep them as they are for now.

devices/Rtc/Devices/Ds1302/Ds1302.cs
```diff
@@ -0,0 +1,186 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Ds1302"/> class.
+        /// </summary>
+        /// <param name="controller">Gpio controller.</param>
+        /// <param name="clockPin">Clock pin number.</param>
+        /// <param name="dataPin">Data pin number.</param>
+        /// <param name="resetPin">Reset pin number.</param>
+        /// <exception cref="ArgumentNullException">When controller was not provided.</exception>
+        public Ds1302(GpioController controller, int clockPin, int dataPin, int resetPin)
```
Ellerbach, line 32: as this one is using GPIO and the other RTC are using I2C, it more than make sense to have this one our of the current nuget and being its own nuget. That will make things lighter for sure.

devices/Rtc/Devices/Ds1302/Ds1302.cs
```diff
@@ -0,0 +1,186 @@
...
+
+namespace Iot.Device.Rtc
+{
+    /// <summary>
+    /// Realtime Clock Ds1302.
+    /// </summary>
+    [Interface("Realtime Clock Ds1302")]
+    public class Ds1302 : RtcBase, IDisposable
+    {
+        private GpioController _controller;
```
Ellerbach, line 18: can be readonly, same for all the GpioPin.

devices/Rtc/Devices/Ds1302/Ds1302.cs
```diff
@@ -0,0 +1,186 @@
...
+        /// </summary>
+        /// <param name="controller">Gpio controller.</param>
+        /// <param name="clockPin">Clock pin number.</param>
+        /// <param name="dataPin">Data pin number.</param>
+        /// <param name="resetPin">Reset pin number.</param>
+        /// <exception cref="ArgumentNullException">When controller was not provided.</exception>
+        public Ds1302(GpioController controller, int clockPin, int dataPin, int resetPin)
+        {
+            _controller = controller ?? throw new ArgumentNullException();
+            _clockPin = _controller?.OpenPin(clockPin, PinMode.Output);
```
Ellerbach, line 35: the controller can't be null here :-) you just checked it before ;-)

Ellerbach: on GpioPin, it's not part of the pattern to use them internally in bindings. But as we are introducing them to .NET IoT, I'm fine to keep them here. Otherwise, just use the pin numbers and it will work the same way

devices/Rtc/Devices/Ds1302/Ds1302.cs
```diff
@@ -0,0 +1,186 @@
...
+            _resetPin = _controller?.OpenPin(resetPin, PinMode.Output);
+            _dataPin = _controller?.OpenPin(dataPin, PinMode.Input);
+
+            _resetPin?.Write(PinValue.Low);
```
Ellerbach, line 39: it can't really be null neither here

devices/Rtc/Devices/Ds1302/Ds1302.cs
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        /// <summary>
+        /// Sets date and time.
+        /// </summary>
+        /// <param name="time">Date and time.</param>
+        protected override void SetTime(DateTime time)
+        {
+            PrepareWrite(Ds1302Registers.REG_WP);
+            WriteByte(0b00000000);
```
Ellerbach, line 92:
```suggestion
            WriteByte(0b0000_0000);
```

It's easier to read. Now, I would ove to have a comment as well from where this magic value is coming from

devices/Rtc/Devices/Ds1302/Ds1302.cs
```diff
@@ -0,0 +1,186 @@
...
+            PrepareWrite(Ds1302Registers.REG_BURST);
+            WriteByte(NumberHelper.Dec2Bcd(time.Second));
+            WriteByte(NumberHelper.Dec2Bcd(time.Minute));
+            WriteByte(NumberHelper.Dec2Bcd(time.Hour));
+            WriteByte(NumberHelper.Dec2Bcd(time.Day));
+            WriteByte(NumberHelper.Dec2Bcd(time.Month));
+            WriteByte(NumberHelper.Dec2Bcd((int)time.DayOfWeek));
+            WriteByte(NumberHelper.Dec2Bcd(time.Year - 2000));
+
+            WriteByte(0b1000000);
```
Ellerbach, line 104:
```suggestion
            WriteByte(0b0100_0000);
```

Same easier to read and please check that it's the proper one as I've adjusted to add a 0 in front to have 2 groups of 4. And as always, a quick comment for the magic number :-)

devices/Rtc/Devices/Ds1302/Ds1302.cs
```diff
@@ -0,0 +1,186 @@
...
+            }
+
+            return readValue;
+        }
+
+        private void PrepareRead(Ds1302Registers ds1302Register)
+        {
+            SetDirectionOfDataPin(PinMode.Output);
+            _resetPin.Write(PinValue.High);
+            WriteByte((byte)(0b10000001 | (byte)ds1302Register));
```
Ellerbach, line 143:
```suggestion
            WriteByte((byte)(0b1000_0001 | (byte)ds1302Register));
```

devices/Rtc/Devices/Ds1302/Ds1302.cs
```diff
@@ -0,0 +1,186 @@
...
+            SetDirectionOfDataPin(PinMode.Input);
+        }
+
+        private void PrepareWrite(Ds1302Registers ds1302Register)
+        {
+            SetDirectionOfDataPin(PinMode.Output);
+            _resetPin.Write(PinValue.High);
+            WriteByte((byte)(0b10000000 | (byte)ds1302Register));
```
Ellerbach, line 151:
```suggestion
            WriteByte((byte)(0b1000_0000 | (byte)ds1302Register));
```

Ellerbach (commented):
Thanks, great quality as always!

devices/Rtc/Devices/Ds1302/Ds1302.cs
rkalwak: As for the source of magic value - no problem, as for formatting - Stylecop forces this as it doesn't know latest language features.

Ellerbach: we do have quite a lot of other cases with the underscore, I'm a bit surprise about that one.

networkfusion:
It also looks like you need to add the following to the `.nuspec` file:

```
nanoFramework.Runtime.Native.1.5.4
nanoFramework.System.Device.Gpio.1.0.4
nanoFramework.System.Device.Model.1.1.113.2032
nanoFramework.System.Diagnostics.Stopwatch.1.1.113.2032
```

devices/Ds1302/Ds1302.cs
```diff
@@ -0,0 +1,186 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Ds1302"/> class.
+        /// </summary>
+        /// <param name="controller">Gpio controller.</param>
+        /// <param name="clockPin">Clock pin number.</param>
+        /// <param name="dataPin">Data pin number.</param>
+        /// <param name="resetPin">Reset pin number.</param>
+        /// <exception cref="ArgumentNullException">When controller was not provided.</exception>
+        public Ds1302(GpioController controller, int clockPin, int dataPin, int resetPin)
```
Ellerbach, line 35: in the pattern, we give as well a _shouldDispose. See the pattern here: https://github.com/nanoframework/nanoFramework.IoT.Device/blob/0d1d97db4da3e306349724d369fd477fa4701055/devices/A4988/A4988%20.cs#L34
And it's used in the dispose as well.
And you should then place the GpioController after the pins and have it with a default value of null. And shouldDispose to be true as default value.

devices/Ds1302/Ds1302.cs
```diff
@@ -0,0 +1,186 @@
...
+        {
+            _controller = controller ?? throw new ArgumentNullException();
+            _clockPin = _controller.OpenPin(clockPin, PinMode.Output);
```
Ellerbach, line 38: you should check that all the pins are valid >=0 (except resetPin, see a bit after). I know, an exception will be thrown by the native part but anyway, best to check here.

devices/Ds1302/Ds1302.cs
```diff
@@ -0,0 +1,186 @@
...
+            EndTransmission();
+        }
+
+        /// <summary>
+        /// Releases the unmanaged resources used by the RtcBase and optionally releases the managed resources.
+        /// </summary>
+        /// <param name="disposing">True to release both managed and unmanaged resources; false to release only unmanaged resources.</param>
+        protected override void Dispose(bool disposing)
+        {
+            _dataPin?.Dispose();
```
Ellerbach, line 117: see the pattern as well with the _shouldDispose. In all cases, you'll have to dispose the GpioPin, and the controller if asked for it. With your current pattern, it's disposed all the time and if it is pass to the binding and used for something else, it won't work properly.

devices/Ds1302/RtcBase.cs
```diff
@@ -0,0 +1,49 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
Ellerbach, line 1: why don't you leave this file in the other binding? and just reference it in the nfproj? That *should* work

Ellerbach (commented):
Thanks, few adjustment needed for the GpioController and _shouldDispose pattern. Rest seems good :-)

devices/Ds1302/RtcBase.cs
rkalwak: Good point, I forgot to revert this...

devices/Ds1302/Ds1302.cs
rkalwak: That is not going to work or would be weird because the RtcBase already implements IDisposable, so adding parameter doesn't sound like a good solution?

devices/Ds1302/Ds1302.cs
```diff
@@ -0,0 +1,199 @@
...
+        /// <param name="resetPin">Reset pin number.</param>
+        /// <param name="controller">Gpio controller.</param>
+        /// <param name="shouldDispose">True to dispose the Gpio Controller.</param>
+        /// <exception cref="ArgumentNullException">When controller was not provided.</exception>
+        public Ds1302(int clockPin, int dataPin, int resetPin, GpioController? controller = null, bool shouldDispose = true)
+        {
+            _controller = controller ?? new GpioController();
+            _shouldDispose = shouldDispose || controller is null;
+
+            if (clockPin < 0 || dataPin < 0 || resetPin < 0)
```
Ellerbach, line 42: reset pin is fine to be negative as you can put it to the ground.

devices/Ds1302/Ds1302.cs
```diff
@@ -0,0 +1,199 @@
...
+            {
+                throw new ArgumentOutOfRangeException();
+            }
+
+            _clockPin = _controller.OpenPin(clockPin, PinMode.Output);
+            _resetPin = _controller.OpenPin(resetPin, PinMode.Output);
+            _dataPin = _controller.OpenPin(dataPin, PinMode.Input);
```
Ellerbach, line 49: check if the pin is positive and open it. See the other comment

devices/Ds1302/Ds1302.cs
```diff
@@ -0,0 +1,199 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Ds1302"/> class.
+        /// </summary>
+        /// <param name="clockPin">Clock pin number.</param>
+        /// <param name="dataPin">Data pin number.</param>
+        /// <param name="resetPin">Reset pin number.</param>
+        /// <param name="controller">Gpio controller.</param>
+        /// <param name="shouldDispose">True to dispose the Gpio Controller.</param>
+        /// <exception cref="ArgumentNullException">When controller was not provided.</exception>
```
Ellerbach, line 36: that's not true anymore ;-) it's when some of the pins are not valid

devices/Ds1302/Ds1302.cs
```diff
@@ -0,0 +1,199 @@
...
+        private readonly bool _shouldDispose;
+        private PinMode _currentDataPinDirection;
+        private GpioController _controller;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Ds1302"/> class.
+        /// </summary>
+        /// <param name="clockPin">Clock pin number.</param>
+        /// <param name="dataPin">Data pin number.</param>
+        /// <param name="resetPin">Reset pin number.</param>
```
Ellerbach, line 33: mention abut negative to be ignored

devices/Ds1302/Ds1302.cs
```diff
@@ -0,0 +1,199 @@
...
+            if (clockPin < 0 || dataPin < 0 || resetPin < 0)
+            {
+                throw new ArgumentOutOfRangeException();
+            }
+
+            _clockPin = _controller.OpenPin(clockPin, PinMode.Output);
+            _resetPin = _controller.OpenPin(resetPin, PinMode.Output);
+            _dataPin = _controller.OpenPin(dataPin, PinMode.Input);
+
+            _resetPin.Write(PinValue.Low);
```
Ellerbach, line 51: and then lower this if the pin is positive only.

so all up, it would look like this:

```csharp
if (clockPin < 0 || dataPin < 0)
{
        throw new ArgumentOutOfRangeException();
 }

 _clockPin = _controller.OpenPin(clockPin, PinMode.Output);
_dataPin = _controller.OpenPin(dataPin, PinMode.Input);

if (resetPin >= 0)
{
    _resetPin = _controller.OpenPin(resetPin, PinMode.Output);
    _resetPin.Write(PinValue.Low);
}
```

devices/Ds1302/Ds1302.cs
```diff
@@ -0,0 +1,199 @@
...
+            EndTransmission();
+        }
+
+        /// <summary>
+        /// <inheritdoc/>
+        /// </summary>
+        /// <param name="disposing">True to release both managed and unmanaged resources; false to release only unmanaged resources.</param>
+        protected override void Dispose(bool disposing)
+        {
+            if (_shouldDispose)
```
Ellerbach, line 126: Make it like this rather because you always need to close the pins but it's only if the controller needs to be disposed that you have to dispose it:

```csharp
            _dataPin?.Dispose();
            _clockPin?.Dispose();
            _resetPin?.Dispose();
            if (_shouldDispose)
            {
                _controller?.Dispose();
                _controller = null;
            }
```

Ellerbach (commented):
Almost there ;-) Few more things for the correct pattern on the dispose and on rest pins ;-)

devices/Ds1302/Ds1302.cs
rkalwak: I beg to differ here, I have no knowledge about eletronics but from logical perspective to make communication successful you have to set the resetPin to be high at some point -> see method PrepareRead. So would logical to have it open first right? The way you are proposing you won't open the pin at all if it was negative (shorted to ground) so in my opinion it suppose to fail on write as pin was not open or simply do not work as intented? 

Having that in mind I suggest to keep it always positive.

Ellerbach: OK. So I checked the datasheet and understand now. The "Reset" name is not appropriate. In old datasheet it was called like this but it's not a reset. Reason why the in name changed for "CE" which should refer to "CommunicationEnabled". So I would suggest then to adjust either the name of the pin, either add a comment about this.

image

Ellerbach (commented):
See comment on the rest pin.
Also, please add the version file and the category file. Then we will be good to go!

devices/Ds1302/version.json
```diff
@@ -0,0 +1,16 @@
+{
+  "$schema": "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
+  "version": "1.2",
```
Ellerbach, line 3: 1.0 as it's a brand new one :-)

devices/Ds1302/RtcBase.cs
Ellerbach: Reopening this comment as the file is back :-)

Ellerbach (commented):
Tiny little things then we're all good!

devices/Ds1302/RtcBase.cs
Ellerbach: If you can't get rid of it, can you please add a comment in the file, right after the header that the original one is in the Rtc folder?

Ellerbach (approved):
Here we are :-) Great addition! Thanks!

PR #740, opened 2023-03-03
Changes for the Adafruilt Feather Wing 128x64 OLED support

jbferguson75:
The issue with the Adafruit specific screen is that natively this screen is 64x128 portrait.  In order to make it work as most intend, the initialization has to be modified (added a 64x128 profile) and a rotation function must be added.  

## Description

- Added initialization profile to support 64x128 pixel screens (as opposed to 128x64).
- Added a Rotation property to set what rotation is desired for the screen.  Since most people do not want a 64x128 screen, the rotation allows the screen to be used any direction desired (0, 90, 180, and 270).
- Modified DrawPixel method to convert pixels to support rotation.

## Motivation and Context

- Adds support for the Adafruit Feather Wing 128x64 OLED screen

## How Has This Been Tested?

This has been tested using the Adafruit HUZZAH32 board using I2C communication to the Adafruit OLED screen.  Other profiles weren't modified so it shouldn't affect previously working screens.  The Rotation defaults to 0 (no rotation) so existing code should still work.

## Types of changes

## Checklist:


devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -311,6 +349,18 @@ public void ClearDirectAligned(int x, int y, int width, int height)
             }
         }
 
+        /// <summary>
+        /// In-place swap of a and b values without the use of a temp variable.
+        /// </summary>
+        /// <param name="a">1st variable to be swapped.</param>
+        /// <param name="b">2nd variable to be swapped.</param>
+        private void Ssd1306Swap(ref int a, ref int b)
```
Ellerbach, line 357: naming wise, why Ssd1306? Is the swap specific? If not, then just call Swap. and if specific, how to swap for the other screens?

devices/Ssd13xx/samples/Program.cs
```diff
@@ -13,14 +13,15 @@
...
 // used for the bus, for example
-Configuration.SetPinFunction(21, DeviceFunction.I2C1_DATA);
-Configuration.SetPinFunction(22, DeviceFunction.I2C1_CLOCK);
+Configuration.SetPinFunction(22, DeviceFunction.I2C1_DATA);
+Configuration.SetPinFunction(20, DeviceFunction.I2C1_CLOCK);
 //////////////////////////////////////////////////////////////////////
 
 //Tested with 128x64 and 128x32 OLEDs
-using Ssd1306 device = new Ssd1306(I2cDevice.Create(new I2cConnectionSettings(1, Ssd1306.DefaultI2cAddress)), Ssd13xx.DisplayResolution.OLED128x64);
+using Ssd1306 device = new Ssd1306(I2cDevice.Create(new I2cConnectionSettings(1, Ssd1306.DefaultI2cAddress)), Ssd13xx.DisplayResolution.OLED64x128);
```
Ellerbach, line 21: Rather than changing this line, add another line and add a comment.

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -319,7 +369,23 @@ public void ClearDirectAligned(int x, int y, int width, int height)
         /// <param name="inverted">Indicates if color to be used turn the pixel on, or leave off.</param>
         public void DrawPixel(int x, int y, bool inverted = true)
         {
-            if ((x >= Width) || (y >= Height))
+            switch (Rotation)
+            {
+                case 1:
```
Ellerbach, line 374: And here, use the enum rather than the numbers.

Ellerbach (commented):
Thanks for the addition. Please use an enum for the rotation.

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -102,6 +130,11 @@ public abstract class Ssd13xx : IDisposable
         /// </summary>
         protected I2cDevice _i2cDevice { get; set; }
 
+        /// <summary>
+        /// Gets or sets Screen rotation.  0 = no rotation, 1 = 90, 2 = 180, 3 = 270.
+        /// </summary>
+        public int Rotation { get; set; } = 0;
```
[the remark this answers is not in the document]
jbferguson75: This is a good idea.  Since it's actually a reference to a rotation rather than an absolute orientation, I changed this to RotationOptions.  If the display is 64x128 then it defaults into portrait with no rotation but if the display is 128x64 than it defaults into landscape with no rotation.  Also these little screens don't necessarily have an up or down so it could be confusing to have a Portrait vs Portrait180 just depending on how you're holding the LCD screen.  I think just having everything with reference to the default orientation is more straightforward.

devices/Ssd13xx/Ssd13xx.cs
jbferguson75: Yeah, I was borrowing code from the AdaFruit C++ driver where it's written this way.  I'll change it to just Swap.

devices/Ssd13xx/samples/Program.cs
jbferguson75: Weird.  I thought I'd left the default one there.  My test changes must have made it in.  I don't think I really need to show the other option in a comment.  I'll just make sure it's still the original line.

devices/Ssd13xx/Ssd13xx.cs
jbferguson75: Done

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -41,6 +41,34 @@ public abstract class Ssd13xx : IDisposable
...
+            0xd5,       // 1 clk div
+            0xae,       // turn display off
+            0xd5, 0x80, // set display clock divide ratio/oscillator,  set ratio = 0x80
+            0xa8, 0x3f, // set multiplex ratio 0x00-0x3f        
+            0xd3, 0x1f, // set display offset 0x00-0x3f, no offset = 0x00
+            0x40 | 0x0, // set display start line 0x40-0x7F
+            0x8d, 0x14, // set charge pump,  enable  = 0x14  disable = 0x10
+            0x20, 0x00, // 0x20 set memory address mode,  0x0 = horizontal addressing mode
+            0xa0 | 0x01, // set segment re-map
+            // 0xa1 | 0x01, // set segment re-map
```
Ellerbach, line 60: if this line is not needed then remove it otherwise please add a little comment on why it should be kept.

devices/Ssd13xx/RotationOptions.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Iot.Device.Ssd13xx
+{
+    /// <summary>
+    /// Options for rotation of the screen.
+    /// </summary>
+    public enum RotationOptions
```
Ellerbach, line 9: Just to be sure why you decided to go for rotation rather than the display orientation: is it because the original display can be different?
And why not remapping with the display orientation I proposed?
I4m asking as we're currently thinking of extracting some core display feature to work with both the pure managed drivers we have (like this one) and the native ones.

devices/Ssd13xx/Ssd13xx.cs
```diff
@@ -319,7 +369,23 @@ public void ClearDirectAligned(int x, int y, int width, int height)
         /// <param name="inverted">Indicates if color to be used turn the pixel on, or leave off.</param>
         public void DrawPixel(int x, int y, bool inverted = true)
         {
-            if ((x >= Width) || (y >= Height))
+            switch (Rotation)
+            {
+                case RotationOptions.Rotate90Degrees:
```
Ellerbach, line 374: you're missing the no rotation case. In case I'm rotating few times, if I then rotate again to no rotation, this may not work properly?

Ellerbach (commented):
Few more questions and then we are good to go! Thanks again!

devices/Ssd13xx/Ssd13xx.cs
jbferguson75: Yeah.  Left there while I was testing.  It's now removed.

devices/Ssd13xx/RotationOptions.cs
jbferguson75: The displays have a default orientation that's sometimes confusing.  Most are oriented to landscape by default but some (like the Adafruit FeatherWing) is oriented in portrait by default.  I could map it to the ones you proposed but I wasn't 100% confident that a screen that's set up as 64x128 is always portrait oriented by default and one that is 128x64 is always oriented landscape by default.  It would make sense, but I wasn't 100% sure.  That's why I opted for the enum that's in reference to whatever the default orientation happens to be.  That way should there be some weird scenario, it would still make sense.  

With all of that being said, I can go by that assumption.  It's an easy change if that's better for other needs.

devices/Ssd13xx/Ssd13xx.cs
jbferguson75: The enum is a property in reference to the default orientation of the display and not a method that rotates it each time it's called.  You set the property once and it's done.  In fact, it may be useful to make it part of the constructor.  

If there is no rotation (default value of the Rotation property) then no byte switching is necessary.  Nothing to be done.

jbferguson75:
@dotnet-policy-service agree

devices/Ssd13xx/RotationOptions.cs
jbferguson75: OK.  It took some figuring out, but it now supports absolute orientations.  To do this, I make the assumption that if the width < height then the default orientation is portrait.  I figure that's a safe assumption.  From there I figure out how to rotate by counting the 90 degree rotations necessary from the assumed default orientation.

jbferguson75: Also found a bug in the text centering when rotating from a default portrait orientation to a landscape orientation.  Fixed.

jbferguson75:
I just realized that the build is failing.  It seems to have an issue with the NanoFramework.System.Math package.  I can remove the one time I use the library but I would think it should work.  Any suggestions on this?  It builds fine on my machine.

devices/Ssd13xx/Ssd13xx.nfproj
```diff
@@ -35,30 +35,32 @@
...
       <HintPath>packages\nanoFramework.System.Device.Gpio.1.1.28\lib\System.Device.Gpio.dll</HintPath>
-      <Private>True</Private>
     </Reference>
-    <Reference Include="System.Device.I2c, Version=1.1.9.0, Culture=neutral, PublicKeyToken=c07d481e9758c731">
+    <Reference Include="System.Device.I2c">
       <HintPath>packages\nanoFramework.System.Device.I2c.1.1.9\lib\System.Device.I2c.dll</HintPath>
-      <Private>True</Private>
+    </Reference>
+    <Reference Include="System.Math">
+      <HintPath>packages\nanoFramework.System.Math.1.5.29\lib\System.Math.dll</HintPath>
```
Ellerbach, line 50: if you are adding math here, you need to add in the nuspec file. Reason why your build is failing.

Ellerbach (commented):
To fix the build, add the missing dependency in the nuspec ;-) Otherwise, looks good :-)

jbferguson75:
Updated the nuspec

devices/Ssd13xx/Ssd13xx.nuspec
```diff
@@ -22,6 +22,7 @@
       <dependency id="nanoFramework.CoreLibrary" version="1.14.2" />
       <dependency id="nanoFramework.System.Device.I2c" version="1.1.9" />
       <dependency id="nanoFramework.System.Device.Gpio" version="1.1.28" />
+	  <dependency id="nanoFramework.System.Math" version="1.5.29" />
```
Ellerbach, line 25:
```suggestion
    <dependency id="nanoFramework.System.Math" version="1.5.29" />
```

Ellerbach (approved):
Tiny tab vs space. Rest looks good! Thanks!

devices/Ssd13xx/Ssd13xx.nuspec
jbferguson75: Good grief.  Forgot to check that.  It's been fixed.

PR #798, opened 2023-05-11
Add support for the Sensirion SEN5x family sensors

martijns:
## Description
* Adds support for the Sensirion SEN5x family devices. This is an all-in-one sensor for PM, temperature, humidity, VOC gasses and NOx gasses.
* Updates the device-listing (for auto-update of README) to include the "nox" category.

## How Has This Been Tested?
The driver was tested on the Sensirion SEN54 variant of the SEN5x series. All functionality could be tested, except the NOx-specific settings which is unique to the SEN55 variant. It was connected to an M5Core2 device on the default I2C pins.

## Types of changes

- [x] Config and build (change in the configuration and build system, has no impact on code or features)

## Checklist:


Ellerbach:
Thanks for the PR, can you add the linter on the project? See https://github.com/nanoframework/nanoFramework.IoT.Device/tree/develop/StyleCop

Your code is already following a great pattern. The linter will find tiny little things, missing intellisense comments, things like this.

devices/Sen5x/Entities/StringValue.cs
```diff
@@ -0,0 +1,47 @@
...
+using System;
+
+namespace Iot.Device.Sen5x.Entities
+{
+    /// <summary>
+    /// Represents a null-terminated string value read from the sensor.
+    /// </summary>
+    internal class StringValue : AbstractReadEntity
+    {
+        private int _maxLength;
```
Ellerbach, line 12: you should be able to have it readonly as well, doesn't seems you're changing the value anywhere (maybe in the underlying class but it doesn't seems so)

devices/Sen5x/Entities/StringValue.cs
```diff
@@ -0,0 +1,47 @@
...
+            // Internal constructor, because this is a read-only entity.
+            _maxLength = maxLength;
+        }
+
+        internal override int ByteCount => _maxLength / 2 * 3;
+
+        internal override void FromSpanByte(SpanByte data)
+        {
+            Text = string.Empty;
+            for (int i = 0; i < data.Length; i += 3)
```
Ellerbach, line 25: for maintainability, can you please add a little comment on the format used by this sensor?
Because all up, seems like a Encoding.Default.GetString will do the job

devices/Sen5x/Entities/VOCAlgorithmState.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// <summary>
+    /// Represents the VOC algorithm state.
+    /// </summary>
+    public class VocAlgorithmState : AbstractReadWriteEntity
+    {
+        internal override int ByteCount => 12;
+
+        internal override void FromSpanByte(SpanByte data)
+        {
+            State = new byte[8];
```
Ellerbach, line 16: Same here and in the various conversion, a comment on how the format looks like would be very appreciated for maintenance point of view.

devices/Sen5x/Extensions/SpanByteExtensions.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// A convenient method that verifies all pairs within a data span. The data must be a multiple of 3 bytes (2 bytes data, 1 byte crc).
+        /// </summary>
+        /// <param name="data">The data to be verified.</param>
+        /// <exception cref="ArgumentException">When the byte lenght is not a multiple of 3.</exception>
+        /// <exception cref="IndexOutOfRangeException">When the CRC fails.</exception>
+        internal static void VerifyCrc(this SpanByte data)
+        {
+            if (data.Length % 3 != 0)
+            {
+                throw new ArgumentException(nameof(data), "Bytes in data must be a multiple of 3");
```
Ellerbach, line 19: you can remove the text of the exception to gain space in the generated PE file. You've already added them in the intellisense comment. We're levering this pattern. So except if the comment is a real must, please remove them from the exceptions in all the project. But make sure you have them in the intellisense comments

devices/Sen5x/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+![Sen5x-image.png](./Sen5x-image.png)
+
+## Documentation
+
+For the implementation of this library the official [datasheet](https://developer.sensirion.com/fileadmin/user_upload/customers/sensirion/Dokumente/15_Environmental_Sensor_Node/Datasheets/Sensirion_Environmental_Sensor_Node_SEN5x_Datasheet.pdf). Any other documentation can be found on its [product page](https://developer.sensirion.com/sensirion-products/sen5x-environmental-sensor-node/).
+
+There are three variants at the time of this writing:
+* `SEN50-SDN-T`, which outputs: Particulate Matter
```
Ellerbach, line 16: Markdown 1:1: you need an extra enter before enumerations
```suggestion

* `SEN50-SDN-T`, which outputs: Particulate Matter
```

devices/Sen5x/Sen5xSensor.cs
```diff
@@ -0,0 +1,355 @@
...
+        {
+            _i2c = i2cDevice;
+        }
+
+        /// <summary>
+        /// Starts the measurement. After power up, the module is in Idle-Mode. Before any measurement values can be read, the Measurement-Mode needs to be started using this command.
+        /// </summary>
+        public void StartMeasurement()
+        {
+            Read(0x0021, TimeSpan.FromMilliseconds(50));
```
Ellerbach, line 44: rather than all those magic numbers, can you create an internal enum with a proper naming for them?
It's much nicer and easier to maintain. It does avoid problems later on when adding a new one for example.
So change as well the signature of your Read function to take this enum and things will be easier to maintain

devices/Sen5x/Sen5xSensor.cs
```diff
@@ -0,0 +1,355 @@
...
+            return (TemperatureCompensationParameters)Read(0x60b2, TimeSpan.FromMilliseconds(20), new TemperatureCompensationParameters());
+        }
+
+        /// <summary>
+        /// These commands allow to compensate temperature effects of the design-in at customer side by applying a custom temperature offset to the ambient temperature.
+        /// </summary>
+        /// <param name="value">The new temperature compensation parameters to be set.</param>
+        public void WriteTemperatureCompensationParameters(TemperatureCompensationParameters value)
+        {
+            Write(0x60b2, TimeSpan.FromMilliseconds(20), value);
```
Ellerbach, line 99: And of course, same remark for the write function and for the read one. Use an internal enum, I guess can be the same one.

devices/Sen5x/Sen5xSensor.cs
```diff
@@ -0,0 +1,355 @@
...
+        }
+
+        private AbstractReadEntity Read(ushort cmd, TimeSpan commandExecutionTime, AbstractReadEntity entity = null)
+        {
+            SpanByte data = new byte[2];
+            BinaryPrimitives.WriteUInt16BigEndian(data, cmd);
+            var result = _i2c.Write(data);
+            if (result.Status != I2cTransferStatus.FullTransfer)
+            {
+                throw new InvalidOperationException($"Unexpected status ({result.Status}) on I2C write");
```
Ellerbach, line 294: note that the exception pattern not to have the text apply here as well. Just extract it in the public functions calling it or just don't do it, there is only 2 functions where this expection is happening. So it's very easy to spot in the code as well.

src/device-listing/Program.cs
```diff
@@ -56,6 +56,7 @@
     "reader",
     "satellite",
     "particulatematter",
+    "nox",
```
Ellerbach, line 59: Can this be just the same category as gas sensor?

devices/Sen5x/category.txt
```diff
@@ -0,0 +1,5 @@
+particulatematter
+hygrometer
+thermometer
+voc
+nox
```
Ellerbach, line 5: see question about using gas sensor.

Ellerbach (commented):
Great addition, very nice and clean, few remarks to make things easier to maintain!

devices/Sen5x/Entities/VOCAlgorithmState.cs
martijns: It feels a bit weird to have to explain the "every 3rd byte is a checksum" even though this is perfectly explained in the datasheet. I don't really document when I use `UInt` vs `Int` either. Since I'm not deviating from the datasheet, it feels redundant to explain. If there's a bug, a developer would have to verify my comments against the official documentation anyway.

I'll add a little note to the abstract methods in the base classes that explain the checksum thing.

devices/Sen5x/Entities/StringValue.cs
martijns: Encoding.UTF8.GetString() won't work (without additional steps). I've added a comment in the code to explain why.

src/device-listing/Program.cs
martijns: I'd indeed favor the existing category. No need to splinter up into even more categories unnecessarily. You could even consider consolidating "voc" into "gas" as well.

martijns:
Thanks for the detailed review. All comments have been addressed/processed.

Ellerbach (approved):
Thanks, great addition!

PR #902, opened 2023-10-23
add GDEW0154x e-paper driver

xas:
## Description

* add a driver for the M5 CoreInk screen
* fix an issue with inherited font size properties

## Motivation and Context

* I have a CoreInk, I want to display data.

## How Has This Been Tested?

* run a sample project to verify it

## Screenshots

CoreInk

## Types of changes

## Checklist:


xas:
Should we separate code into 3 package nuget ?

* EPaper
* GDEW0154x
* SSD168x

(on draft state, as I am waiting some specs confirmation from M5 team)

devices/ePaper/Drivers/Gdew0154x/Gdew0154x.cs
```diff
@@ -0,0 +1,636 @@
...
+            int busyPin,
+            int dataCommandPin,
+            int width,
+            int height,
+            GpioController gpioController,
+            bool enableFramePaging = false)
+        {
+            _spiDevice = spiDevice ?? throw new ArgumentNullException(nameof(spiDevice));
+            _gpioController = gpioController ?? throw new ArgumentNullException(nameof(gpioController));
+
```
Ellerbach, line 63: can you add a "_shouldDispose" to check if the gpio controller should be disposed? See the pattern here for example: https://github.com/nanoframework/nanoFramework.IoT.Device/blob/9ce639dc2f38b94ffe1f2cfb6606a180b97bfb23/devices/A4988/A4988%20.cs#L20
In short, you can pass a null GpioController and you'll create it, also you'll dispose it in this case. If it's not null and the shouldDispose is true, you'll dispose it otherwise no.

devices/ePaper/Drivers/Gdew0154x/Gdew0154x.cs
```diff
@@ -0,0 +1,636 @@
...
+        /// </summary>
+        /// <param name="buffer">The buffer array to draw.</param>
+        public void DirectDrawBuffer(params byte[] buffer)
+        {
+            SendCommand((byte)Command.DataStartTransmission2);
+            SendData(buffer);
+            WaitMs(5);
+
+            SendCommand((byte)Command.DisplayRefresh);
+            WaitMs(500);
```
Ellerbach, line 217: can you add a little comment on why to wait 5 ms and then 500 ms? Just a comment like "// 500 ms needed as per documentation page 5" will do it. So it makes it easier when maintaining and debugging all this

devices/ePaper/Drivers/Gdew0154x/Gdew0154x.cs
```diff
@@ -0,0 +1,636 @@
...
+        /// <exception cref="InvalidOperationException">Unable to initialize the display until it has been powered on. Call PowerOn() first.</exception>
+        public virtual void Initialize()
+        {
+            if (PowerState != PowerState.PoweredOn)
+            {
+                throw new InvalidOperationException();
+            }
+
+            SendCommand((byte)Command.PanelSetting);
+            SendData(0xdf);
```
Ellerbach, line 392: can you point out where the initilization sequence is coming form? So it makes it easier for maintenance

devices/ePaper/Drivers/Gdew0154x/Gdew0154x.cs
```diff
@@ -0,0 +1,636 @@
...
+                return true;
+            }
+
+            return false;
+        }
+
+        /// <inheritdoc/>
+        public virtual void PerformFullRefresh()
+        {
+            WaitMs(10);
```
Ellerbach, line 458: why 10 milliseconds?

devices/ePaper/Drivers/Gdew0154x/Gdew0154x.cs
```diff
@@ -0,0 +1,636 @@
...
+            PowerState = PowerState.PoweredOff;
+        }
+
+        /// <summary>
+        /// Performs the required steps to "power on" the display.
+        /// </summary>
+        public virtual void PowerOn()
+        {
+            HardwareReset();
+            WaitMs(1000);
```
Ellerbach, line 496: few comments on the sequence will be greatly appreciated as well

devices/ePaper/Drivers/Gdew0154x/Gdew0154x.cs
```diff
@@ -0,0 +1,636 @@
...
+            PowerState = PowerState.PoweredOn;
+        }
+
+        /// <inheritdoc/>
+        public virtual void SendCommand(params byte[] command)
+        {
+            // make sure we are setting data/command pin to low (command mode)
+            _dataCommandPin.Write(PinValue.Low);
+
+            foreach (var b in command)
```
Ellerbach, line 511: maybe for performance reasons, it could be faster to write all the bytes at once? As every time, the data is sent independently with a low and high on the chip select. So it's much slower than sending in a single write with a buffer.

devices/ePaper/Drivers/Gdew0154x/Gdew0154x.cs
```diff
@@ -0,0 +1,636 @@
...
+            }
+        }
+
+        /// <inheritdoc/>
+        public virtual void SendData(params byte[] data)
+        {
+            // set the data/command pin to high to indicate to the display we will be sending data
+            _dataCommandPin.Write(PinValue.High);
+
+            foreach (var @byte in data)
```
Ellerbach, line 524: same comment as in the command one

devices/ePaper/Drivers/Gdew0154x/Gdew0154x.cs
```diff
@@ -0,0 +1,636 @@
...
+        {
+            Thread.Sleep(milliseconds);
+        }
+
+        /// <inheritdoc/>
+        public virtual void WaitReady(PinValue pinValue)
+        {
+            while (_busyPin.Read() != pinValue)
+            {
+                WaitMs(5);
```
Ellerbach, line 584: if waiting is a matter of few milliseconds you can use the Thread.Spin. If it's indeed much more, then keep the Thread.Sleep

devices/ePaper/Drivers/Gdew0154x/Gdew0154x.cs
```diff
@@ -0,0 +1,636 @@
...
+            }
+        }
+
+        /// <inheritdoc/>
+        public virtual void WaitReady()
+        {
+            while (_busyPin.Read() == PinValue.Low)
+            {
+                WaitMs(5);
```
Ellerbach, line 593: same remark here as well

Ellerbach (commented):
Amazing start! And I love the fact that the interface makes sense :-) that's the first real different driver added here! So it's great to see! Few comments mainly about getting clarity for maintainability.
Also, you'll need to bump the minor version in the version file.

Ellerbach:
> Should we separate code into 3 package nuget ?

I guess it's ok to have one only for the moment (it simplify a bit things), the code overhead is not that high

devices/ePaper/Drivers/Gdew0154x/Gdew0154x.cs
xas: I did, but I think there are some non-sense between samples I found on M5 repo and the one provided by the screen manufacturer...

devices/ePaper/Drivers/Gdew0154x/Gdew0154x.cs
xas: I did a little rewrite for the WaitBusy function as it was waiting indefinitely

devices/ePaper/Drivers/Ssd168x/Ssd168x.cs
```diff
@@ -627,12 +628,16 @@ protected virtual void WaitMs(int milliseconds)
         }
 
         /// <inheritdoc/>
-        public virtual void WaitReady()
+        public virtual bool WaitReady(int waitingTime)
         {
-            while (_busyPin.Read() == PinValue.High)
+            int currentWait = 0;
+            while (currentWait < waitingTime && _busyPin.Read() == PinValue.High)
```
Ellerbach, line 634: create a cancellation token, it's done for that. And you can use it in the loop as well to be non blocking

devices/ePaper/Drivers/Gdew0154x/Gdew0154x.cs
```diff
@@ -0,0 +1,661 @@
...
+        /// <exception cref="ArgumentNullException"><paramref name="spiDevice"/> is null.</exception>
+        /// <exception cref="ArgumentOutOfRangeException">Display width and height can't be less than 0 or greater than 200.</exception>
+        protected Gdew0154x(
+            SpiDevice spiDevice,
+            int resetPin,
+            int busyPin,
+            int dataCommandPin,
+            int width,
+            int height,
+            GpioController gpioController,
```
Ellerbach, line 62: you can even set this one as null and create it internally if not given.

Ellerbach (commented):
better and better ;-)

MrCSharp22 (changes requested):

xas:
For the busy pin question, I am more on the solution of warn the developer, but not block him
The current implementation of `WaitReady` is a blocking thread, so if something happen on the screen side, the all app is blocked.

I think it should be the developer role to decide what to do if after a while the busy pin is still up.

I am also against raising an exception as you could crash the developer's app.

I am more on a warning policy: "hey be careful, seems there are still work on the screen, but continue with the rest"

xas:
BTW, is it really necessary `Iot.Device.EPaper.Drivers.JD79653A.JD79653A` ?

I don't think a class and a namespace should share a same name. Why the class `Iot.Device.EPaper.Drivers.JD79653A`should be wrong ?

Ellerbach:
> BTW, is it really necessary `Iot.Device.EPaper.Drivers.JD79653A.JD79653A` ?

no it's not. Fine to use directly the Driver namespace!

devices/ePaper/Drivers/JD796xx/JD79653A.cs
```diff
@@ -0,0 +1,664 @@
...
+using Iot.Device.EPaper.Buffers;
+using Iot.Device.EPaper.Enums;
+using nanoFramework.UI;
+
+namespace Iot.Device.EPaper.Drivers.JD796xx
+{
+    /// <summary>
+    /// Base class for GDEW0154-based ePaper devices.
+    /// </summary>
+    public abstract class JD79653A : IEPaperDisplay
```
Ellerbach, line 18: Technically the name should be `Jd79653A` to follow our guidelines. Same for the namespace `Jd796xx`

devices/ePaper/Drivers/Ssd168x/Ssd168x.cs
```diff
@@ -18,6 +18,7 @@ namespace Iot.Device.EPaper.Drivers.Ssd168x
     /// </summary>
     public abstract class Ssd168x : IEPaperDisplay
     {
+        private readonly int _maxWaitingTime = 500;
```
Ellerbach, line 21: why not a const?

Ellerbach (commented):
Couple of small details. Great to see the PR now real review :-)

devices/ePaper/Drivers/Ssd168x/Ssd168x.cs
xas: I want to leave the possibility that become a configurable property when instantiating it.

devices/ePaper/Drivers/Ssd168x/Ssd168x.cs
Ellerbach: Something like (just did it from memory) this is much cleaner and you won't have to wait an extra 5 ms (which is ultimately going to be 20 ms) unnecessary. this shows as well how to set a default time.

```csharp
public virtual bool WaitReady(CancellationToken waitingTime = default)
{
  if (waitingTime == default)
  {
    // With a default time of 5 seconds as an example
    waitingTime = new CancellationTokenSource(5000);
  }

  while(!waitingTime.IsCancelledRequest && _busyPin.Read() == PinValue.High)
  {
    waitingTime.WaitHandle.WaitOne(5, true);
  }

  return !waitingTime.IsCancelledRequest;
}
```

Ellerbach (commented):
Thanks, just a little detail on the wait function using cancellation token

Ellerbach (approved):
Looks all good now! Thanks!

Ellerbach (approved):
I think we're good to go :-) HUGE thanks for the work on this contribution! Great to see this additional epaper screen!

PR #966, opened 2024-01-05
Add color transition and asynchronous methods to RgbDiode class

torbacz:
## Description
This update enhances the RgbDiode class with methods for transitioning between colors smoothly, both synchronously and asynchronously. New methods include color transitioning on a separate thread, fade transitions, and smooth blinking. Also, the 'System.Threading' namespace has been added for support with these new features.

## Types of changes

## Checklist:


devices/RgbDiode/RgbDiode.cs
```diff
@@ -72,6 +80,92 @@ public void SetColor(Color color)
...
+        
+        /// <summary>
+        /// Transitions the pixel to new color in a separate thread.
+        /// </summary>
+        /// <param name="color">The new color.</param>
+        /// <param name="cancellationToken">Cancellation token.</param>
+        /// <param name="steps">Number of transition steps.</param>
+        /// <param name="delay">Delay between each transition step.</param>
+        /// <returns>The new Thread or null if color is the same as current color.</returns>
+        public Thread TransitionAsync(Color color, CancellationToken cancellationToken, int steps = DefaultSteps, int delay = DefaultDelay)
```
Ellerbach, line 94: why do you return the thread? Would it be better to return a cancellation token or equivalent rather than the thread itself?
Also any sample on how to properly use this new feature? :-)

torbacz: Well, I've implemented it based on "Task" approach from full .NET (maybe that's wrong approach for Threads?). But, in my opinion token is something that we are providing from outside of method - to be able to cancel thread. 
I think, I should add default to CancellationToken.

As for sample, I'll provide one in readme and sample soon.

torbacz: Included sample

Ellerbach (approved):
OK with Thread, right, similar to Task and when we'll have those, it can then be replaced like this. So looks good to me now. Thanks again.

Ellerbach:
@torbacz note I'll wait for the nuget issue to be resolved to merge it.

PR #973, opened 2024-01-24
[Ws28xx.Esp32] Change visibility of several classes and properties

sangyuxiaowu:
## Description

- Made the class base `BitmapImage` class public
- Made the `ClockDivider`, `OnePulse`, `ZeroPulse`, and `ResetCommand` parameters in `ws28xx` public.

## Motivation and Context

I encountered difficulties when using the Iot.Device.Ws28xx.Esp32 library due to my WS2812B device possibly being non-standard, with the model number XL-0807RGBC-WS2812B. When attempting to control it, I found that the Ws2815b in Iot.Device.Ws28xx.Esp32 could not drive it, while Ws2815c could, but the colors were reversed.

The README suggests inheriting from the Ws28xx class and making necessary modifications. However, the BitmapImageWs2808Grb and BitmapImageWs2808 classes, which could otherwise be used directly, are internal and thus cannot be inherited. Even though the source code could be inherited directly from BitmapImage, the parameters within Ws28xx, namely ClockDivider, OnePulse, ZeroPulse, and ResetCommand, are also internal and cannot be inherited. This means that I would need to rewrite all these parameters myself, which greatly reduces the usefulness of the library.

My motivation for this change is to make these classes and parameters public, allowing users to inherit them and modify as needed, enhancing the library's flexibility and usability.

- Resolves nanoframework/Home#1433

## Types of changes

## Checklist:

### DIFF
--- a/devices/Ws28xx.Esp32/BitmapImageNeo3.cs
+++ b/devices/Ws28xx.Esp32/BitmapImageNeo3.cs
@@ -10,7 +10,7 @@ namespace Iot.Device.Ws28xx.Esp32
     /// Special 24bit RGB format for Neo pixel LEDs where each bit is converted to 3 bits.
     /// A one is converted to 110, a zero is converted to 100.
     /// </summary>
-    internal class BitmapImageNeo3 : BitmapImage
+    public class BitmapImageNeo3 : BitmapImage
     {
         protected const int BytesPerComponent = 3;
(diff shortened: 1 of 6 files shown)


Ellerbach:
Looks good but please fix the errors from the linter and we will be good to go. So we can merge on green! Thanks!

Ellerbach (approved):
Thanks a lot, now I hope this will allow more scenarios as you've presented!

PR #978, opened 2024-01-30
[RgbDiode] Add missing chanel start. Update sample

torbacz:
## Description

- Add missing "channel.Start" for each ADC channel
- Adjust sample code + adjust sample code in readme

## Motivation and Context

- After upgrading to latest FW on ESP32 lib was not working. I've added missing Start method for each channel
- I've found a bug - looks like bug in ADC handling on native side (I don't own logic analyser so can't confirm). For now created workaround and left a comment.

## How Has This Been Tested?

- ESP32 REV0 with latest firmware

## Types of changes

## Checklist:

### DIFF
--- a/devices/RgbDiode/README.md
+++ b/devices/RgbDiode/README.md
@@ -20,8 +20,8 @@ using Iot.Device.RgbDiode;
 using System.Drawing;
 
 const byte redGpioPin = 25;
-const byte greenGpioPin = 26;
-const byte blueGpioPin = 27;
+const byte greenGpioPin = 27;
+const byte blueGpioPin = 26;
 
 // Uncomment for ESP32
 // nanoFramework.Hardware.Esp32.Configuration.SetPinFunction(redGpioPin, nanoFramework.Hardware.Esp32.DeviceFunction.PWM1);
--- a/devices/RgbDiode/README.md
+++ b/devices/RgbDiode/README.md
@@ -17,15 +17,35 @@ The binding supports the following sensor functions:
 
 ```csharp
 using Iot.Device.RgbDiode;
+using System.Drawing;
 
 const byte redGpioPin = 25;
-const byte greenGpioPin = 27;
-const byte blueGpioPin = 26;
+const byte greenGpioPin = 26;
+const byte blueGpioPin = 27;
 
-var pwm = new RgbDiode(redGpioPin, greenGpioPin, blueGpioPin);
+// Uncomment for ESP32
+// nanoFramework.Hardware.Esp32.Configuration.SetPinFunction(redGpioPin, nanoFramework.Hardware.Esp32.DeviceFunction.PWM1);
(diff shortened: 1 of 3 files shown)


Ellerbach (approved):
That will indeed fix the bug found by one of our Discord member and good improvement for the invert color

torbacz:
@Ellerbach 
That was me on Discord :D 
Yet this is more like workaround, rather than fix.

As for the invert part, it's inverting only PWM channel in case you are using common anode (invert = true) or common cathode (invert = false) diode. Maybe I should rename it? Or create enum?

Ellerbach:
> Yet this is more like workaround, rather than fix.

@alberk8 maybe something you want to look at as you've been improving and fixing PWM recently. See more context in Discord and the fact you have to set 2 times the duty channel to change.

> As for the invert part, it's inverting only PWM channel in case you are using common anode (invert = true) or common cathode (invert = false) diode. Maybe I should rename it? Or create enum?

Yes, keep it for the next improvement. I was not shocked by the name.

alberk8:
Frankly speaking. I would want a  `Public boot IsStarted => _isStarted` to indicate that the PWM has started and so that the Frequency and DutyCycle does not to be called multiple times.   The present API allows you to set the two properties without starting the PWM and those call goes directly to native.   Also the [PwmPulsePolarity](https://github.com/nanoframework/System.Device.Pwm/blob/abbd3ebdb71baf893646e5249ed9bda1baf5bb21/System.Device.Pwm/PwmChannel.cs#L26) is never exposed.  

I have not sure what are the implication on other platforms if we change this.

Ellerbach:
> Frankly speaking. I would want a `Public boot IsStarted => _isStarted` to indicate that the PWM has started and so that the Frequency and DutyCycle does not to be called multiple times [here](https://github.com/nanoframework/System.Device.Pwm/blob/abbd3ebdb71baf893646e5249ed9bda1baf5bb21/System.Device.Pwm/PwmChannel.cs#L164) [here](https://github.com/nanoframework/System.Device.Pwm/blob/abbd3ebdb71baf893646e5249ed9bda1baf5bb21/System.Device.Pwm/PwmChannel.cs#L45)

We can adjust the API, that's not a big problem. The goal of our API is to be aligned with .NET IoT: https://github.com/dotnet/iot/blob/main/src/System.Device.Gpio/System/Device/Pwm/PwmChannel.cs

Depending on the API, there has been small divergences, also to take care of the history (like for I2C), the way things were done in nanoFramework.

Adding a IsStarted could be done easilly. It's something we can even bring back to .NET IoT. As an example GpioPin was not exisitng in .NET IoT and it's something that was brought back from nanoFramework.

Feel free to propose a PR on PWM. Keep in mind, that the signature will change, meaning, you also need to adjust the nf-interpreter side. And the IsStarted internal variable could be set either natively (maybe the best?) or on managed code. In that case, it's easier and less changes natively.

PR #1088, opened 2024-07-17
ATGM336H GPS module

torbacz:
## Description

- Implementation of ATGM336H GPS module.

## Motivation and Context

- New binding

## How Has This Been Tested?

- ESP32 with ATGM336H module

## Screenshots
image

## Types of changes

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Introduced categorization for devices with new entries for "gps" and "helper."
  - Added versioning and build configurations for both Atgm336h and GpsDevice.

- **Documentation**
  - New metadata files for device categorization and versioning management.

- **Chores**
  - Enhanced structured dependency management for the .NET nanoFramework across multiple devices.


devices/Atgm336h/Atgm336h.cs
```diff
@@ -0,0 +1,233 @@
...
+            _serialDevice = new SerialPort(serialPort);
+            _serialDevice.BaudRate = 9600;
+            _serialDevice.Parity = Parity.None;
+            _serialDevice.StopBits = StopBits.One;
+            _serialDevice.Handshake = Handshake.None;
+            _serialDevice.DataBits = 8;
+            _serialDevice.ReadBufferSize = 2048;
+            _serialDevice.ReadTimeout = 2000;
+            _serialDevice.WatchChar = '\r';
+            _serialDevice.DataReceived += SerialDevice_DataReceived;
```
Ellerbach, line 91: open question: do you want to add the event before starting? And remove them after stopping?

devices/Atgm336h/Atgm336h.cs
```diff
@@ -0,0 +1,233 @@
...
+            _shouldDispose = false;
+        }
+
+        /// <summary>
+        /// Starts the GPS module by opening the serial device connection.
+        /// </summary>
+        public void Start()
+        {
+            _serialDevice.Open();
```
Ellerbach, line 111: I would feel better to check if it's already started and skip opening it a second time. Something simple with a private variable.

devices/Atgm336h/Atgm336h.cs
```diff
@@ -0,0 +1,233 @@
...
+        }
+
+        /// <summary>
+        /// Stops the communication with the GPS module by closing the serial port.
+        /// </summary>
+        public void Stop()
+        {
+            _serialDevice.Close();
```
Ellerbach, line 119: and same here, closing only if it's open to avoid problems.

devices/Atgm336h/Atgm336h.cs
```diff
@@ -0,0 +1,233 @@
...
+            if (serialDevice.BytesToRead == 0)
+            {
+                return;
+            }
+
+            var buffer = new byte[serialDevice.BytesToRead];
+            var bytesRead = serialDevice.Read(buffer, 0, buffer.Length);
+            var stringBuffer = Encoding.UTF8.GetString(buffer, 0, bytesRead);
+            var commands = stringBuffer.Split(serialDevice.WatchChar);
+            foreach (var command in commands)
```
Ellerbach, line 179: thinking load here: this is most likely not the only device with GPS we will add.
Would it make sense to have a minimal nuget parsing simple standard GPS with a bit of extensibility?
There are some elements on the .NET IoT repository: https://github.com/dotnet/iot/tree/main/src/devices/Common/Iot/Device/Common
And a very (too complete) implementation for nano as well in .NET IoT: https://github.com/dotnet/iot/tree/main/src/devices/Nmea0183
Other interest of separating this are the ability to add quite some unit tests and have a simple, low memory, efficient and robust parser.
We can start small and produce a nuget for a very minimal implementation as a start, taking this part, maybe some of the functions you have, adding a simple way to extend. Very open for the discussion.

devices/Atgm336h/Atgm336h.cs
```diff
@@ -0,0 +1,233 @@
...
+            {
```
Ellerbach, line 180: as you are manipulating strigs, that then may not be complete, I would go for a try/catch for this function.

devices/Atgm336h/Atgm336h.cs
```diff
@@ -0,0 +1,233 @@
...
+        {
+            if (_serialDevice is null)
+            {
+                return;
+            }
+            
+            if (_shouldDispose)
+            {
+                _serialDevice.Dispose();
+                _serialDevice.DataReceived -= SerialDevice_DataReceived;
```
Ellerbach, line 228: I would remove the event before disposing

devices/Atgm336h/README.md
```diff
@@ -0,0 +1,58 @@
+# GPS Module
```
Ellerbach, line 1:
```suggestion
# ATGM336H  - Serial GPS Module

```

devices/Atgm336h/README.md
```diff
@@ -0,0 +1,58 @@
...
+Configuration.SetPinFunction(txPin, DeviceFunction.COM2_TX);
+var gpsModule = new Atgm336h(comPort);
+gpsModule.LocationChanged += GpsModuleOnLocationChanged;
+gpsModule.FixChanged += GpsModuleOnFixChanged;
+gpsModule.ModeChanged += GpsModuleOnModeChanged;
+gpsModule.Start();
+
+Thread.Sleep(Timeout.Infinite);
+
+void GpsModuleOnFixChanged(Fix fix)
```
Ellerbach, line 44: and another question for a longer discussion: why not having an abstract class as well in a simple nuget with the generic parser that already implement functions like this one?
All GPS have something similar. In short, we would have a basic parser, the base class with the event, the fix, the modes, all this. And then the ATGM336H would just have to implement what is specific.

Ellerbach (commented):
All up, all good,. I think, if you're ok with the principle, we can do better here in splitting this binding in 2:
* 1 with a core parser, a GPS abstract class (not necessary serial, just with the core functions), the enums, things like this. All are common to all GPS.
* 1 with what's specific.

I understand it's a bit more work but I think it can bring a lot of value as I don't think we have any other GPS.

devices/Atgm336h/Atgm336h.cs
torbacz: Is there any difference? It should not be invoked without start. But maybe it should be moved into Open method?
So add after start and remove before stop?

devices/Atgm336h/Atgm336h.cs
torbacz: Sure, why not. Just point me to repository and I'll create something :)

devices/Atgm336h/Atgm336h.cs
torbacz: 🤦‍♂️ 
Yep, you totally right.

devices/Atgm336h/README.md
torbacz: Abstract class or maybe static helper? Like WeatherHelper class (maybe it should be named GpsHelper)?

devices/Atgm336h/Atgm336h.cs
Ellerbach: yes, I think it's better to add an remove at start and stop, things stays clean then

devices/Atgm336h/Atgm336h.cs
Ellerbach: Well, we can have it in the IoT repository :-) So create a folder, create another nuget, project reference is fine for this one and any other that will use it. The nuget is a bous for anyone who wants to use it externally.

devices/Atgm336h/README.md
Ellerbach: I think all GPS are quire similar. And the even they can raise are similar as well. So having an abstract class at minima, with possibly static helpers can be a great combination.
I guess the idea is to keep it nano compatible, simple and efficient. The .NET IoT repo code is too complex for nano. But it has the advantage of containing all the methos. Also, it has tests and other elements that can be reused.

devices/Atgm336h/Atgm336h.cs
torbacz: done

devices/Atgm336h/Atgm336h.cs
torbacz: done

devices/Atgm336h/Atgm336h.cs
torbacz: done

devices/Atgm336h/Atgm336h.cs
torbacz: done

devices/Atgm336h/Atgm336h.cs
torbacz: done + added exception event

torbacz:
Left todo:

- [ ] Create base class for all GPS
- [ ] Create static helper for GPS parsing

devices/Atgm336h/README.md
torbacz: done

devices/Atgm336h/Atgm336h.cs
torbacz: done

torbacz:
I had to remove StyleCop from "GpsDevice" project - no idea why it doesn't downlaod nuget package when there is a package reference.

torbacz:
Something wrong with reference to other project, probalby packages path :/

torbacz:
Waiting for #1100

Ellerbach:
So, now I look back at his PR, in regard of the GnssDevice work, we can definitely create a fully generic GNSS Receiver or equivalent. All commands are fully standard, the logic should be the same for all classic receviers. They can have additional commands both in receive and send mode as well. So, we can improve the GnssDevice binding to add a GenericGnssRecevier (or equivalent) which will implement all the classical elements with serial support.

Ellerbach:
@torbacz shall we close this one as your case is covered by the generic one?
You may add it as an example and mention the name of the device (I'll add the ones I4ve been tested) in the description, so we can increase the referencing. What do you think?

torbacz:
@Ellerbach 
The way I look for supported devices is to check list here:
https://github.com/nanoframework/nanoFramework.IoT.Device

And I think a lot of users are doing the same - so as long as we can add it to this list without binding itself - I'm ok with it

Ellerbach:
> And I think a lot of users are doing the same - so as long as we can add it to this list without binding itself - I'm ok with it

I've added into the description. But there are thousands of GPS supported. So we won't create thousands of nuget which does the exact same thing. I'm happy to add few names into the title of the binding but definitely not creating packages that are not really necessary

torbacz:
@Ellerbach Makes sens. Closing.

PR #1100, opened 2024-07-27
Generic GnssDevice

torbacz:
## Description

- New project for all GPS devices across NF project.

## Motivation and Context

- As the build with project reference failes to build on pipeline I decieded to split it into 2 separate PRs and use package reference in GPS module.

## Types of changes

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced a categorization system for GPS and GNSS devices, enhancing organization and interaction.
	- Implemented structured dependency management for both GPS and GNSS devices, ensuring consistent builds.
	- Established versioning and release management configurations for both GPS and GNSS projects to improve deployment processes.


devices/GpsDevice/NMEA0183Parser.cs
```diff
@@ -0,0 +1,90 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace Iot.Device.Common.GpsDevice
+{
+    /// <summary>
+    /// Provides methods for parsing NMEA0183 data from a GPS device.
+    /// </summary>
+    public static class NMEA0183Parser
```
Ellerbach, line 11: Question: shall we follow our naming rules and use Nmea0183Parser instead?

devices/GpsDevice/NMEA0183Parser.cs
```diff
@@ -0,0 +1,90 @@
...
+    {
+        /// <summary>
+        /// Parses the GNGLL NMEA0183 data from a GPS device.
+        /// </summary>
+        /// <param name="inputData">The valid GNGLL data string starting with $GNGLL.</param>
+        /// <returns>The parsed GngllData object.</returns>
+        public static GngllData ParaseGngll(string inputData)
```
Ellerbach, line 18: as all the NMEA string are fixed, can you create a kind of generic parser?
Something that can give one of the GngllData or GngsaData file?
I would see a generalized based class names "NmeaData" which contains a field with the signature.
As an example, it could look like:

```csharp
public interface NmeaData
{
    public string Id {get; }
    public NmeaData Parse(string inputData);
}

public class GngsaData: NmeaData
{
    public override string Id = "$GNGSA";
   public NmeaData Parse(string inputData);
    // Rest of the class
}

// And in the parser

public static class NMEA0183Parser
{
    public NmeaParser[] MneaDatas {get; protected set; }

      // In the constructor, just add all the NmeaData derivated class that are supported
     //  MneaDatas = new MneaData[] { GngsaData }; // add all supported

    public NmeaData Parse(string inputData)
    {
        var data = inputData.Split(',');
        for each (string id in NmeaDatas)
        {
            if (id.Id == data[0])
            {
                return id.Parse(inputData);
            }
        }
    }
}
```

Advantage: you can add later on any parser you'd like!

devices/GpsDevice/category.txt
```diff
@@ -0,0 +1,2 @@
+gps
```
Ellerbach, line 1: not sure we have those categories yet, please check the project that builds automatically all the readmes! If needed, can you add those into the categoies? (sorry, a bit in a rush to check into details)

Ellerbach (commented):
I love it! See one of my comment to make it extendible easily!

networkfusion:
Would it not  be better to call this the GnssDevice base project to better represent that it can be used for Galilao etc.?

Ellerbach:
> Would it not be better to call this the GnssDevice base project to better represent that it can be used for Galilao etc.?

Yes, good catch btw! That's indeed a great idea. We can later on add the elements. Reminds me I've been working on GNSS for the SIM Com modems. Changes are still in a branch (which is btw ready for PR): https://github.com/nanoframework/nanoFramework.IoT.Device/tree/add-atmodem-gnss/devices/AtModem

So most likely elements to scaange there!

networkfusion:
> > Would it not be better to call this the GnssDevice base project to better represent that it can be used for Galilao etc.?
> 
> Yes, good catch btw! That's indeed a great idea. We can later on add the elements. Reminds me I've been working on GNSS for the SIM Com modems. Changes are still in a branch (which is btw ready for PR): https://github.com/nanoframework/nanoFramework.IoT.Device/tree/add-atmodem-gnss/devices/AtModem
> 
> So most likely elements to scaange there!

I also want to draw attention to the https://learn.microsoft.com/en-us/dotnet/maui/platform-integration/device/geolocation?view=net-maui-8.0&tabs=windows namespace...

Ellerbach:
> Maybe we can hijack the API and rename the namespace to nanoFramework...
> If the API is sound, of course, which I haven't looked into yet. 

So I had a look at all this and I don't think it's relevant for a device. Basically what's there is a service to get a geolocalisation. Which needs to use a device, totally hidden in that case. Here, it's kind of the opposite. BUT, we can try to reuse the same names for the Localisation and few others liks the events.

@torbacz I don't have much time but I think I can help a bit with this PR so that we can use it both for your other PR and the SIMCom modem. So that all classes, base class, enums, namespaces are aligned. Let me know if that's ok for you, so I can work on a this PR as well!

torbacz:
@Ellerbach Sure, feel free to commit :)

devices/GnssDevice/Nmea0183Parser.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            return null;
+        }
+
+        /// <summary>
+        /// Parses the GNGLL NMEA0183 data from a Gnss device.
+        /// </summary>
+        /// <param name="inputData">The valid GNGLL data string starting with $GNGLL.</param>
+        /// <returns>The parsed GngllData object.</returns>
+        public static GngllData ParaseGngll(string inputData)
```
torbacz, line 42: @Ellerbach Should we remove it from parser?

devices/GnssDevice/Nmea0183Parser.cs
```diff
@@ -0,0 +1,114 @@
...
+{
+    /// <summary>
+    /// Provides methods for parsing NMEA0183 data from a Gnss device.
+    /// </summary>
+    public static class Nmea0183Parser
+    {
+        /// <summary>
+        /// Gets or sets a list of NMEA0183 data objects.
+        /// </summary>
+        public static INmeaData[] MneaDatas { get; set; }
```
torbacz, line 16: How this will be populated?

Ellerbach: I just realized I missed a push :-( All is here now

devices/GnssDevice/Nmea0183Parser.cs
Ellerbach: yes, yes, again, sorry, miss to push a commit yesterday, please check now, should be all good.

Ellerbach:
@torbacz I think I fixed all the core elements, including the tests (you can't pass enums directly, so need to be casted). See the tests which are now running with nano.
Things to be added:
* Extract a timestamp for the commands an add it to the GeoLocalisation
* Depending on the commands, populate more elements (like altitude but also the precision)
* Add a README
* Add few more very used commands?

torbacz:
@Ellerbach 
I've pushed some changes:
- Changed how MneaDatas are handled (Hashtable instead of collection)
- Changed Nmea0183Parser.Parse return type to object (this interface won't work in my opinion as not all data will have the same fields - this way it's much safer in my opinion. BTW I really miss generic, this is a perfect case for them :( )
- Added some "Debug" error handling 
- Some docs fixes and changed visibility of methods

Let me know If you are ok with such version.

As for TODOs, I would go with minimal version right now. Then we can add more/adjust existing commands.

Ellerbach:
@torbacz 
> Changed Nmea0183Parser.Parse return type to object (this interface won't work in my opinion as not all data will have the same fields

as you've simplified the interface and anyway only classes with this interface can be processed, better return the strong type

for the rest, added functions to add or remove from the collection, so it's a bit safer

and added time and date management

There is a test to fix (you created it originally) and then we can indeed go with this first version

torbacz:
@Ellerbach 
Test fixed, so I think we are ready to merge this version?

Ellerbach (approved):
Let's go for this first version. I'll most likely add more elements in another PR. I'm happy with the current implementation to start with! Thanks again @torbacz for the great collaboration.

PR #1217, opened 2025-01-03
[AtModem] Fix Uncaught exception in ModemBase.GetNetworkRegistration()

rikkreeftenberg:
## Description
After sending "AT+CREG?" the answer sometimes doesn't contain a ','.

Like in:
 • +CREG: 1,0
 • +CREG: 1,0
 • +CREG: 1
 • +CREG: 1,1

Then an uncaught exception is thrown because of the line:

                    int stat = int.Parse(parts[x]);

But this doesn't happen very often. Only once every 10 networkregistrations  or so…

The fix is to check the lenght before asuming there are at least 2 parts.

## Motivation and Context
- This bug sometimes causes Uncaught Exceptions. It happens only every 10 NetworkRegistrations or so.
- When the system reboots (after an Error or a watchdog trigger) the NetworkRegistration is performed again.
- When this gives an Uncaught exception the results are possibly a hanging system.
- This fix makes the system more stable.

## How Has This Been Tested?
- It has been tested by running the system for days while it occasionally reboots. Everytime it needs to call GetNetworkRegistration(). The output of the system is monitored by Putty and written to logfiles.
These logfiles don't show this particular Uncaught Exception anymore.

## Types of changes

## Checklist:

### DIFF
--- a/devices/AtModem/Modem/ModemBase.cs
+++ b/devices/AtModem/Modem/ModemBase.cs
@@ -722,7 +722,7 @@ namespace Iot.Device.AtModem.Modem
 
                     // The first one is the notification level with proactive +CREG
                     ////int n = int.Parse(parts[0]);
-                    int stat = int.Parse(parts[1]);
+                    int stat = parts.Length >= 2 ? int.Parse(parts[1]) : 0;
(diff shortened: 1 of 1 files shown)


Ellerbach (approved):
Thanks as well for catching up all those little bugs!

PR #1218, opened 2025-01-03
[AtModem] Fix system hang in sim7672 HTTP client send

rikkreeftenberg:
## Description
When the response is processed at the end of Sim7672HttpClient.SendInternal the idea was to limit the time to 2 minutes.

Because of a logical bug this time is unlimited so when something unexpected happens and there remain bytes to read the system continues to wait indefinitely.

The while loop should read: 
While there is something left to read AND Cancelation is NOT requested
DO

Also, when the line simply reads "ERROR" then you can stop reading and break the While loop. Otherwise it will wait for 2 minutes. (Or indefinitely if the previous bug is not fixed)

This bug happens very few times. Once every couple of hours. So it's very difficult to reporoduce..
But when it does, you see the following in the logfile:
In: 
In: 
In: 
In: 
In: 
In: 
In: 

## Motivation and Context
Fixing this bug restores the way this functionality was supposed to work.
It improves the stability of the library.
Catching the 'ERROR' line and breaking immediately is a small improvement of the flow.

## How Has This Been Tested?

This has been tested by sending lots of http POST messages for multiple days and logging the output with Putty.
In the resulting Logfiles the hanging situation doesn't occur anymore.

## Types of changes

- [] Improvement (non-breaking change that improves a feature, code or algorithm)
- [] New feature (non-breaking change which adds functionality to code)
- [] Breaking change (fix or feature that would cause existing functionality to change)
- [] Config and build (change in the configuration and build system, has no impact on code or features)
- [] Dependencies (update dependencies and changes associated, has no impact on code or features)
- [] Unit Tests (add new Unit Test(s) or improved existing one(s), has no impact on code or features)
- [] Documentation (changes or updates in the documentation, has no impact on code or features)

## Checklist:

- [] My code follows the code style of this project (only if there are changes in source code).
- [] My changes require an update to the documentation (there are changes that require the docs website to be updated).
- [] I have updated the documentation accordingly (the changes require an update on the docs in this repo).

### DIFF
--- a/devices/AtModem/Http/Sim7672HttpClient.cs
+++ b/devices/AtModem/Http/Sim7672HttpClient.cs
@@ -249,18 +249,18 @@ namespace Iot.Device.AtModem.Http
                     Modem.Channel.Start();
                 }
 
-                Modem.GenericEvent += ModemGenericEvent;
-                _httpActionArrived.Reset();
+                    Modem.GenericEvent += ModemGenericEvent;
+                    _httpActionArrived.Reset();
 
-                Modem.Channel.SendCommand($"AT+HTTPACTION={method}");
+                    Modem.Channel.SendCommand($"AT+HTTPACTION={method}");
 
-                // this is the timeout for the loop to complete, it has to depend on the content length, with a minimum of 15 seconds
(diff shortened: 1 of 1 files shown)


Ellerbach (approved):
And thanks as well!

PR #1219, opened 2025-01-03
[AtModem] Fix HTTP init retry mechanism SIM7672

rikkreeftenberg:
## Description

The first thing that is necessary to send a HTTP message is to send the "AT+HTTPINIT" command. In the current situation when this fails, it will try it one more time and when that fails it just continues as if nothing is wrong..

The unused "Retry:" and the "int retries = 5" variable suggest that there once was a retry mechanism and in the Sim800HttpClient there is.

So I copied this retry mechanism to the Sim7672HttpClient.SendInternal, including the return of "ServiceUnavailable" when it fails after 5 retries.

I haven't noticed this gone wrong but the original implementation suggests it might. And after this fix it is handled correctly.
So it improves the stabillity of the system. And the consistence.

Signed-off-by: Rik Kreeftenberg <rik.kreeftenberg@spinsoft.nl>

## Motivation and Context

This fix is required to handle possible fails to HTTPINIT the modem correctly and in line with the SIM800 modem.
Is it an improvement or a fix? I say it's a bug because only part of the retry mechanism is implemented and when the original retry fails, the system just continues and won't work correctly.

(mind the pattern to link there as all issues are tracked in the Home repository) -->

## How Has This Been Tested?

This has been tested by sending thousends of http messages to an Azure Function app for days on end and monitoring the output with Putty and logfiles. 
To be honest I haven't seen the Retry system kick in and saving the day, but that may happen some times. 

## Types of changes

## Checklist:

### DIFF
--- a/devices/AtModem/Http/Sim7672HttpClient.cs
+++ b/devices/AtModem/Http/Sim7672HttpClient.cs
@@ -90,15 +90,23 @@ namespace Iot.Device.AtModem.Http
                 int index;
                 int retries = 5;
 
+            Retry:
                 // We init, if already init, try to deiniti and reinit
                 response = Modem.Channel.SendCommand($"AT+HTTPINIT");
                 if (!response.Success)
                 {
-                    Modem.Channel.SendCommand($"AT+HTTPTERM");
-                    Modem.Channel.SendCommand($"AT+HTTPINIT");
+                    // Give it another try
+                    if (retries-- > 0)
+                    {
+                        // Experience shows that 1 second is ok in most of the times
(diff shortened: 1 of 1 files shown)


Ellerbach (approved):
Looks good, still for this one, can the sleep be reduced? 1 second is very long.

rikkreeftenberg:
@Ellerbach Sure I can reduce it. Problem is I can't test this because I've never seen this go wrong.
Shall I set it to 250 msec?

Ellerbach (approved):
Thanks, I think it's a good compromize!

PR #1407, opened 2025-08-04
Driver DRV8825 for stepper motor was added

Naglfario:
## Description
- New Drv8825 driver

## Motivation and Context

DRV8825 is a stepper motor driver that is made as compatible as possible with A4988, but there are some nuances. I tried to use the A4988 library to work with my DRV8825 driver, but the driver did not work correctly. After little investigation, I realized that despite the same connectors, this driver is more capricious to the control signal. Thus, I wrote a library to control the stepper motor through the DRV8825 driver.

## How Has This Been Tested?

When I written a new library, I tested it several times through a sample project with my ESP32 WROOM(rev 3), DRV8825 and Nema 17 (stepping motor) to make sure that the motor does not miss steps.

## Types of changes

## Checklist:


devices/Drv8825/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Connections
+
+VMOT - connect to 8-35V (motor supply voltage)
+
+GND - connect to GND
+
+1A, 1B, 2A, 2B - connect to the 4 coils of motor
+
+DIR - connect to microcontroller pin
```
Workshopshed: Would be good to expand this to include a simple description. Perhaps even from the data sheet.

DIR - Connect to microcontroller output pin, sets the direction of stepping. Internal pulldown

STEP - Connect to microcontroller output pin, rising edge causes the indexer to move one step. Internal pulldown.

RST - Connect to microcontroller output pin, active-low, reinitializes the indexer logic and disables H-bridge outputs. Internal pulldown.

SLP - Can be joned to RST or left high to enable the device. Low puts the device to sleep.

devices/Drv8825/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+STEP - connect to microcontroller pin
+
+RST � join with SLP and connect to microcontroller pin or to 3.3-5V
+
+Logic power supply can be connected to FAULT pin, it will works, but it useless.
```
Workshopshed: FAULT - Logic low when in fault condition (overtemp, overcurrent). Connect to Microcontroller input with pullup.

devices/Drv8825/Drv8825_circuit_bb.jpeg
Workshopshed, line 1: I don't believe connecting FAULT to Vcc is the correct thing to do according to the datasheet that's an output which is pulled low for fault conditions. So if there is a fault connecting it to Vcc will cause it to short circuit the powersupply.

devices/Drv8825/README.md
Naglfario: Thank you for you review! I'll work on it later and i'll improve documentation

devices/Drv8825/Drv8825_circuit_bb.jpeg
Naglfario: Well, I understand your concern. However, according to the datasheet: 

> Note that the carrier includes a 1.5k protection resistor in series with the FAULT pin that makes it is safe to connect this pin directly to a logic voltage supply, as might happen if you use this board in a system designed for the pin-compatible A4988 carrier

Workshopshed: So that is specific to the module you are using rather than the driver chip itself. The link below has a simpler circuit where FAULT is not connected which might be a better option, or perhaps the high level wiring diagram on the TI datasheet.

.https://www.pololu.com/product/2133

devices/Drv8825/Drv8825.cs
```diff
@@ -0,0 +1,194 @@
...
+        private readonly GpioPin _stepPin;
+        private readonly GpioPin _dirPin;
+        private readonly GpioPin? _sleepPin;
+        private readonly bool _shouldDisposeGpioController;
+        private readonly MicrostepsController _microstepsController;
+
+        private GpioController? _gpioController;
+
+        /// <summary>
+        /// Gets or sets the delay that allows the driver to recognize a step. By default, it is 5 MICROseconds.
```
Ellerbach:
```suggestion
        /// Gets or sets the delay that allows the driver to recognize a step. By default, it is 5 microseconds.
```

devices/Drv8825/Drv8825.nfproj
```diff
@@ -0,0 +1,73 @@
...
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="packages.config" />
+    <Compile Include="Direction.cs" />
+    <Compile Include="Drv8825.cs" />
+    <Compile Include="StepSize.cs" />
+    <Compile Include="MicrostepsController.cs" />
+    <None Include="README.md" />
+    <Compile Include="Properties\AssemblyInfo.cs" />
+    <None Include="*.md" />
```
Ellerbach: I don't think this is needed as you already have the markdown listed an no other md file.

```suggestion
```

devices/Drv8825/README.md
```diff
@@ -0,0 +1,108 @@
...
+Logic low when driver in fault condition(overtemp, overcurrent). Not necessary connection.
+
+M0, M1, M2 � can be connected to microcontroller output pins to use microsteps,
+it takes smoother movement of motor, but it not necessary connections.
+
+![circuit](./Drv8825_circuit_bb.jpeg)
+
+### Advices:
+
+If you have DRV8825 module(not only chip), it is pin-compatible A4988 carrier.
```
Ellerbach:
```suggestion
If you have DRV8825 module (not only chip), it is pin-compatible A4988 carrier.
```

devices/Drv8825/README.md
```diff
@@ -0,0 +1,108 @@
...
+It means for example that you can connect logic power supply to FAULT pin.
+It will works for DRV8825 module(not just chip!), but it useless. If possible, try to avoid it.
```
Ellerbach:
```suggestion
It will work for DRV8825 module (not just the chip!), but it is useless. If possible, try to avoid it.
```

Ellerbach:
azp /run

Naglfario:
> couple of comments, looks good overall!

@Ellerbach Thank you for your review! I made all the suggested changes manually and pushed them in one commit, because if I accepted each suggestion separately, each of them would result in a new commit, and I don't think that's a good idea. Anyway, how do you do it in nanoFramework? I'll keep that in mind for future reference.

Ellerbach:
> I made all the suggested changes manually and pushed them in one commit, because if I accepted each suggestion separately, each of them would result in a new commi

When you are in the "file changed" tab on GitHub, you can batch accept all the changes, or add changes to the batch and commit once only!

devices/Drv8825/samples/Program.cs
```diff
@@ -0,0 +1,49 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Iot.Device.Drv8825;
+using System.Threading;
+
+const byte stepPin = 26;
```
Ellerbach: minor as for the sample, we use capital letters for the constants.

Naglfario:
> When you are in the "file changed" tab on GitHub, you can batch accept all the changes, or add changes to the batch and commit once only!

I tried to do this, but "Add suggestion to batch" button was not available even in the files tab. I don't know why. But using this method was my first thought.

devices/Drv8825/MicrostepsController.cs
```diff
@@ -0,0 +1,134 @@
...
+                    UpdateM1(PinValue.Low);
+                    UpdateM2(PinValue.High);
+                    break;
+            }
+        }
+
+        /// <summary>
+        /// Sets the value of M0.
+        /// </summary>
+        /// <param name="value">Pin value(Low or High).</param>
```
Ellerbach:
```suggestion
        /// <param name="value">Pin value (Low or High).</param>
```

devices/Drv8825/MicrostepsController.cs
```diff
@@ -0,0 +1,134 @@
...
+                return;
+            }
+
+            _m0Pin.Write(value);
+        }
+
+        /// <summary>
+        /// Sets the value of M1.
+        /// </summary>
+        /// <param name="value">Pin value(Low or High).</param>
```
Ellerbach:
```suggestion
        /// <param name="value">Pin value (Low or High).</param>
```

devices/Drv8825/MicrostepsController.cs
```diff
@@ -0,0 +1,134 @@
...
+                return;
+            }
+            
+            _m1Pin.Write(value);
+        }
+
+        /// <summary>
+        /// Sets the value of M2.
+        /// </summary>
+        /// <param name="value">Pin value(Low or High).</param>
```
Ellerbach:
```suggestion
        /// <param name="value">Pin value (Low or High).</param>
```

devices/Drv8825/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+FAULT � can be connected to microcontroller input pin with pullup.
+Logic low when driver in fault condition(overtemp, overcurrent). Not necessary connection.
+
+M0, M1, M2 � can be connected to microcontroller output pins to use microsteps,
+it takes smoother movement of motor, but it not necessary connections.
+
+![circuit](./Drv8825_circuit_bb.jpeg)
+
+### Advices:
```
Ellerbach:
```suggestion
### Advices
```

Naglfario:
So, again, the batch suggestion feature is not available for me, that's why I do it manually:

image

Ellerbach (approved):
Thanks a lot, looks good!

PR #1516, opened 2026-04-17
Add SX1262 LoRa driver and sample

garydyksman:
## Description
- Added a new `Iot.Device.LoRa` library for LoRa radio support.
- Added an `ILoRaDevice` base interface to provide a common abstraction for LoRa radio implementations.
- Added an `Sx1262` driver implementation for SPI-based SX1262 LoRa transceivers.
- Added an `Sx1262Sample` nanoFramework sample application showing initialization, transmit, and receive polling.
- Fixed the SX1262 `ReadBuffer` SPI read handling so the driver skips the protocol bytes correctly and returns only the received payload.
- Cleaned up minor formatting issues in assembly info files.

## Motivation and Context
- This change adds LoRa support for SX1262-based devices in the `devices` folder and provides a sample to validate and demonstrate usage.
- This change also fixes RX buffer handling in the SX1262 driver so received payloads are read correctly.

## How Has This Been Tested?
- Tested on hardware using the included `Sx1262Sample` application.
- Verified device startup and initialization through debug output.
- Verified RX polling started successfully.
- Verified transmit messages were sent from the sample application.
- Verified packets were received successfully and reported RSSI/SNR values in the debug output.
- Verified the `ReadBuffer` fix by confirming that received payloads were returned correctly in the sample output.
- Note: the `01/01/1970` timestamps in the debug output are expected for this test setup because the device clock was not set.

## Screenshots
- Debug output from `Sx1262Sample` showing successful startup, initialization, RX polling, TX messages, and received packets with RSSI/SNR values.

LoRa_debug_SC

## Types of changes

## Checklist:


Ellerbach:
@garydyksman thanks a lot for this nice addition. Before going for a deeper code review, can you please add StyleCop? See https://github.com/nanoframework/nanoFramework.IoT.Device/tree/main/StyleCop

A lot of things will be catch up by it. I4ll then do a proper review.

Ellerbach:
Also you are missing nuget and few other things. Check the other devices.

garydyksman:
@Ellerbach Hi Laurent, I have the PR in a better state now.
Take a look at the ILoRaDevice interface, let me know if it makes sense, or if i have missed something.

devices/LoRa/LoRa/Drivers/Sx1262.cs
```diff
@@ -0,0 +1,709 @@
...
+        // ---------------------------------------------------------------
+        // Static helpers (SA1204: public static before non-static public)
+        //// ---------------------------------------------------------------
+
+        /// <summary>
+        /// Decodes chip mode bits [6:4] from a raw status byte.
+        /// </summary>
+        /// <param name="status">The status byte returned by the chip.</param>
+        /// <returns>A short label describing the chip mode.</returns>
+        public static string DecodeChipMode(byte status)
```
Ellerbach, line 96: just cosmetic: you can probably use the new way of using switch to return the value, that will be more readable

devices/LoRa/LoRa/Drivers/Sx1262.cs
```diff
@@ -0,0 +1,709 @@
...
+                dio1PinObj = _gpio.OpenPin(dio1Pin, PinMode.Input);
+                resetPinObj.Write(PinValue.High);
+                _resetPin = resetPinObj;
+                _busyPin = busyPinObj;
+                _dio1Pin = dio1PinObj;
+                resetPinObj = null;
+                busyPinObj = null;
+                dio1PinObj = null;
+            }
+            catch
```
Ellerbach, line 177: why not just testing the validity of the pins?

devices/LoRa/LoRa/Drivers/Sx1262.cs
```diff
@@ -0,0 +1,709 @@
...
+            try
+            {
+                resetPinObj = _gpio.OpenPin(resetPin, PinMode.Output);
+                busyPinObj = _gpio.OpenPin(busyPin, PinMode.Input);
+                dio1PinObj = _gpio.OpenPin(dio1Pin, PinMode.Input);
+                resetPinObj.Write(PinValue.High);
+                _resetPin = resetPinObj;
+                _busyPin = busyPinObj;
+                _dio1Pin = dio1PinObj;
+                resetPinObj = null;
```
Ellerbach, line 173: why do you set them to null in here? You just tried to set them up before?

devices/LoRa/LoRa/Drivers/Sx1262.cs
```diff
@@ -0,0 +1,709 @@
...
+        /// <param name="timeoutMs">Maximum time to wait, in milliseconds.</param>
+        public void WaitBusy(int timeoutMs)
+        {
+            int elapsed = 0;
+            while (_busyPin.Read() == PinValue.High)
+            {
+                Thread.Sleep(1);
+                if (++elapsed >= timeoutMs)
+                {
+                    throw new TimeoutException("SX1262 BUSY timeout");
```
Ellerbach: we try to limit the strings put in the code. And use a pattern where we document in the intellisense the exception. Especially when this one is the only, nothing that is output really help and easy to trace back!

devices/LoRa/LoRa/Drivers/Sx1262.cs
```diff
@@ -0,0 +1,709 @@
...
+        }
+
+        // ---------------------------------------------------------------
+        // Step 2 — Full init sequence
+        //// ---------------------------------------------------------------
+
+        /// <inheritdoc/>
+        public void Initialize()
+        {
+            WriteCommand(OpSetDio3AsTcxoCtrl, new byte[] { 0x02, 0x00, 0x01, 0x40 });
```
Ellerbach, line 268: Can you point out where in the docs are all the magic numbers? It will help for the maintenance

devices/LoRa/LoRa/Drivers/Sx1262.cs
```diff
@@ -0,0 +1,709 @@
...
+            WriteCommand(OpSetBufferBaseAddr, new byte[] { 0x00, 0x00 });
+            byte[] irqParams = new byte[] { 0x02, 0x03, 0x02, 0x03, 0x00, 0x00, 0x00, 0x00 };
+            WriteCommand(OpSetDioIrqParams, irqParams);
+        }
+
+        /// <inheritdoc/>
+        public void SetRfFrequency(uint frequencyHz)
+        {
+            ulong frf = ((ulong)frequencyHz << 25) / 32000000UL;
+            byte[] rfFreqBytes = new byte[]
```
Ellerbach: minor: we do have binary converters for all this

devices/LoRa/LoRa/Drivers/Sx1262.cs
```diff
@@ -0,0 +1,709 @@
...
+        {
+            // Poll-thread check must run before _sendLock: the poll thread must never block on _sendLock
+            // (e.g. if a PacketReceived handler calls Send), or it can deadlock with the thread holding the lock.
+            bool wasPolling;
+            lock (_pollLock)
+            {
+                if (_pollThread != null && Thread.CurrentThread == _pollThread)
+                {
+                    throw new InvalidOperationException(
+                        "Send cannot be called from the RX polling thread. Do not call Send from PacketReceived; post work to another thread instead.");
```
Ellerbach: same here for the exception, move the text to the intellisense. It's also an easy one to track with the trace.

devices/LoRa/LoRa/Drivers/Sx1262.cs
```diff
@@ -0,0 +1,709 @@
...
+            restoreRxAfterSend = false;
+
+            if (payload == null)
+            {
+                throw new ArgumentNullException(nameof(payload));
+            }
+
+            if (payload.Length == 0)
+            {
+                throw new ArgumentException("Payload cannot be empty", nameof(payload));
```
Ellerbach: same here for all the exception. You can keep the nameof as it will distinguish which one is faulty.

devices/LoRa/LoRa/Drivers/Sx1262.cs
```diff
@@ -0,0 +1,709 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Reads the chip status byte.
+        /// </summary>
+        /// <returns>The second byte of the status SPI transaction.</returns>
+        public byte GetStatus()
+        {
+            WaitBusy(5000);
```
Ellerbach: I see you start everywhere by waiting. Why not making the array operation first as it's taking some time already and wait right before the transfer?
Apply this everywhere and you'll gain performances!

devices/LoRa/LoRa/Drivers/Sx1262.cs
```diff
@@ -0,0 +1,709 @@
...
+
+            if (wasPolling)
+            {
+                StopPolling();
+                restoreRxAfterSend = true;
+            }
+
+            ClearIrqStatus(0xFFFF);
+
+            byte[] packetParams = new byte[]
```
Ellerbach, line 382: for thos magic numbers, would really be great to have a pointer from the doc

devices/README.md
```diff
@@ -243,6 +243,7 @@
 
 * [![NuGet](https://img.shields.io/nuget/v/nanoFramework.Iot.Device.AtModem.svg?label=NuGet&style=flat&logo=nuget)](https://www.nuget.org/packages/nanoFramework.Iot.Device.AtModem/) [Generic AT Modem SIM800 and SIM7070, SIM7080, SIM7090 - Dual Mode Wireless Module CatM, LTE modems](AtModem)
 * [![NuGet](https://img.shields.io/nuget/v/nanoFramework.Iot.Device.Nrf24l01.svg?label=NuGet&style=flat&logo=nuget)](https://www.nuget.org/packages/nanoFramework.Iot.Device.Nrf24l01/) [nRF24L01 - Single Chip 2.4 GHz Transceiver](Nrf24l01)
+* [![NuGet](https://img.shields.io/nuget/v/nanoFramework.Iot.Device.LoRa.svg?label=NuGet&style=flat&logo=nuget)](https://www.nuget.org/packages/nanoFramework.Iot.Device.LoRa/) [LoRa - Semtech SX1262 transceiver (Iot.Device.LoRa)](LoRa)
```
Ellerbach: that's not needed :-) it's added automatically  at merge :-)

Ellerbach: But you are missing the category.txt file. Se https://github.com/nanoframework/nanoFramework.IoT.Device/blob/main/devices/A4988/category.txt
And for the list, pick existing ones: https://github.com/nanoframework/nanoFramework.IoT.Device/blob/main/src/device-listing/categories.json
If you really want a new one, add it to the list.

Ellerbach (commented):
Thanks, really cool! Couple of comments and suggestions to improve few things.

devices/LoRa/LoRa/Drivers/Sx1262.cs
garydyksman: @Ellerbach  I may be overcomplicating this, so I’d appreciate your take. The intent of the try/catch is: if OpenPin fails partway through, we dispose any pins we already opened so the GPIO controller doesn’t keep them reserved—making a retry in the same session (or tests) safer. I’m not assuming the app always recovers without a reboot on real hardware.

devices/LoRa/LoRa/Drivers/Sx1262.cs
garydyksman: @Ellerbach I have tried to make it prettier, but StyleCop is blocking me from going 'modern'

devices/LoRa/LoRa/Drivers/Sx1262.cs
garydyksman: @Ellerbach i have updated the README.md with link to the datasheet, but there is also a link in the file.

devices/README.md
garydyksman: @Ellerbach i have the README and category file in the root of the project, is that perhaps the wrong place to have the files? Where should they be?

Ellerbach:
Please remove the pdf, we are not checking in datasheet. Most are proprietary of their owner. But you can point out where it is located on the device producer.

devices/LoRa/LoRa/Drivers/Sx1262.cs
Ellerbach: haaa, right, that's more than possible :-)

devices/LoRa/LoRa/Drivers/Sx1262.cs
Ellerbach: ok but you should not set them to null in the try part!
and rather than checking if the pins are null, then call Dispose() (you may have to adjust it a bit for the stoppolling thread)

Ellerbach (commented):
Thanks, just a change needed in the constructor to avoid the pins open to get disposed by the garbage collector in the try, removing the pdf file and we'll be good to go!

PR #1608, opened 2026-06-27
Add Home Assistant integration

Ellerbach:
## Description

Add Home Assistant integration

## Motivation and Context

- Add Home Assistant integration
- Make it simple
- Make it transparent

## How Has This Been Tested?

On real device and in a real project (the sample)

## Types of changes

## Checklist:

- [x] My changes require an update to the documentation (there are changes that require the docs website to be updated).
- [x] I have added new tests to cover my changes.


devices/HomeAssistant/MiniJson.cs
```diff
@@ -0,0 +1,213 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
josesimoes, line 1: Anything preventing from using the full json lib? Just asking.

josesimoes (commented):
Awesome! 
Single observation: this deserves a repo on its own! 😉

josesimoes (approved):

Ellerbach:
> Single observation: this deserves a repo on its own! 😉

Less repos = best ;-) It's a binding, indeed with a different namespace but that's it ;-) It is still relatively simple compare to some of the devices we have here and no native dependency!

devices/HomeAssistant/MiniJson.cs
Ellerbach: Performances as it's deterministic, it's faster like this. And as everything is lowercase, it will then make the decoration and the overhead quite heavy. And also, there are quite some cases where something is needed, not on other places, so, the models will be more complex and will have anyway to have some specific logic. So, after debating with myself and making some tests, this seems easier.

PR #1649, opened 2026-07-21
[HomeAssistant] Don't require double device IDs and auto generate LWT topic

WilliamBZA:
## Description

- Remove the need for a device ID to be passed in to both the `HomeAssistantClient` and the `HomeAssistantInfoDevice` classes.
  - Now the device has an ID that the `HomeAssistantClient` uses as well.
  - This device ID is sanitized to make sure it uses only valid characters for MQTT topics.
 - Auto-generate the LWT used for the availability topic unless it is overriden by the user.
   - The topic used for the LWT is an implementation detail and most users won't need to override it.

## Motivation and Context

These are improvements to the API that users will use to register devices in HomeAssistant. A single Device ID is sufficient for topic generation, so the one that was provided to the `HomeAssistantClient` made the API uncertain. (E.g. I found myself asking which device ID is going to be used for the topics?).

The second change was to auto-generate the topic used for the LWT availability topic. Most users won't need to specify the topic used so auto-generating it makes the API surface simpler. For those users that DO have a valid need to specify the topic, the parameter is now optional so they can still do that if they need.

## How Has This Been Tested?

- Created test devices and added them to my home assistant instance
- Migrated a device that used the previous `nf.HomeAssistant` nuget package

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)
  - This results in a breaking change to the API surface

## Checklist:

### DIFF
--- a/devices/HomeAssistant/HomeAssistantClient.cs
+++ b/devices/HomeAssistant/HomeAssistantClient.cs
@@ -212,7 +212,7 @@ namespace nanoFramework.HomeAssistant
                     sanitized.Append(c);
                     lastWasSeparator = false;
                 }
-                else if (c == ' ' || c == '-' || c == '_')
+                else if (IsAsciiWhitespace(c) || c == '-' || c == '_')
                 {
                     if (!lastWasSeparator && sanitized.Length > 0)
                     {
@@ -230,6 +230,18 @@ namespace nanoFramework.HomeAssistant
             }
 
             return sanitized.ToString();
+        }
+
+        /// <summary>
+        /// Determines whether <paramref name="c"/> is an ASCII whitespace character (space, tab, newline,
+        /// carriage return, form feed, or vertical tab). nanoFramework's mscorlib does not provide
+        /// <c>Char.IsWhiteSpace</c>, so this enumerates the characters explicitly.
+        /// </summary>
+        /// <param name="c">The character to test.</param>
+        /// <returns><see langword="true"/> if <paramref name="c"/> is ASCII whitespace; otherwise <see langword="false"/>.</returns>
+        private static bool IsAsciiWhitespace(char c)
+        {
+            return c == ' ' || c == '\t' || c == '\n' || c == '\r' || c == '\f' || c == '\v';
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


devices/HomeAssistant/HomeAssistantClient.cs
```diff
@@ -88,14 +88,24 @@ public HomeAssistantClient(
...
         {
-            _deviceName = deviceName;
+            if (device == null)
+            {
+                throw new ArgumentNullException(nameof(device));
+            }
+
+            if (SanitizeTopicSegment(device.Id, '-').Length == 0)
+            {
+                throw new ArgumentException("Device ID must contain at least one alphanumeric character (a-z, 0-9) to generate MQTT topics.", nameof(device));
```
Ellerbach: can you move the text from this exception to the intellisense to gain some space? it's something that is purely at developer point of view

devices/HomeAssistant/HomeAssistantDeviceInfo.cs
```diff
@@ -24,65 +21,41 @@ public sealed class HomeAssistantDeviceInfo
         /// <exception cref="ArgumentException">Thrown when <paramref name="id"/> or <paramref name="name"/> is null or empty.</exception>
         public HomeAssistantDeviceInfo(string id, string name, string model = null, string manufacturer = null)
         {
+            if (string.IsNullOrEmpty(id))
+            {
+                throw new ArgumentException("Device id cannot be null or empty.");
```
Ellerbach: same here, move the text to the intellisense

devices/HomeAssistant/HomeAssistantDeviceInfo.cs
```diff
@@ -24,65 +21,41 @@ public sealed class HomeAssistantDeviceInfo
...
+            }
+
+            if (string.IsNullOrEmpty(name))
+            {
+                throw new ArgumentException("Device name cannot be null or empty.");
```
Ellerbach: and here

Ellerbach (commented):
Thanks, looks good! Small improvements (I know, there were some of those exception texts in the code).

Ellerbach (approved):
And approving anyway as those as just nits. And thanks again for the improvements!

devices/HomeAssistant/HomeAssistantClient.cs
WilliamBZA: I've updated, is this what you meant?

